PR #22, opened 2015-03-27
Fix multinomial emission probabilities

sagunb:
Multinomial emission probabilities should be normalised per state.


superbobry:
Looks good, thanks!

Issue #1, opened 2014-03-23
Misc complaints


GaelVaroquaux:
https://github.com/scikit-learn/scikit-learn/issues/1862

* [X] The function normalize returns A / Asum, should do A /= Asum; return A.
* [X] The convergence check in BaseHMM.fit should probably not use abs().
* [X] GaussianHMM._init should probably use the all the observations instead of just the first one to compute the means and covariances.
* [X] Would be nice if BaseHMM.fit would set an attribute self.logprob_ = logprob before returning self, so that callers can examine how the training went.

PR #23, opened 2015-03-27
Add requested attribute from issue #1

sagunb:
Add two extra attributes: n_iterations_performed and logprob_delta to allow users to gain insight into the convergence behaviour

### DIFF
--- a/hmmlearn/base.py
+++ b/hmmlearn/base.py
@@ -68,6 +68,17 @@ class _BaseHMM(BaseEstimator):
         startprob, 't' for transmat, and other characters for
         subclass-specific emmission parameters. Defaults to all
         parameters.
+        
+    n_iter_performed : int
+        Number of iterations performed while training the model.
+        
+    logprob_ : float
+        The log probability of the model after training has completed.
+        
+    logprob_delta : float
+        The last update to the model's log probability before training
+        was terminated. If this parameter is negative, or is less than
+        the convergence threshold, the model did not converge.
 
     See Also
     --------
--- a/hmmlearn/base.py
+++ b/hmmlearn/base.py
@@ -103,7 +103,7 @@ class _BaseHMM(BaseEstimator):
         self.algorithm = algorithm
         self.random_state = random_state
         self.logprob_ = None
-        self.n_iterations_performed = None
+        self.n_iter_performed_ = None
         self.logprob_delta = None
 
     def eval(self, X):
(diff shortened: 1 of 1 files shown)


hmmlearn/base.py
```diff
@@ -102,6 +102,9 @@ def __init__(self, n_components=1, startprob=None, transmat=None,
         self.transmat_prior = transmat_prior
         self.algorithm = algorithm
         self.random_state = random_state
+        self.logprob_ = None
+        self.n_iterations_performed = None
```
superbobry: This is simply `len(self.logprob_)`, right?

hmmlearn/base.py
```diff
@@ -102,6 +102,9 @@ def __init__(self, n_components=1, startprob=None, transmat=None,
...
+        self.logprob_delta = None
```
superbobry: I'm not sure we want this as an extra attribute, because the deltas can be trivially computed from `len(self.logprob_)`.

hmmlearn/base.py
sagunb: n_iterations_performed would not be len(self.logprob_). In self.logprob_ I am only saving the last value of the log prob (it is not an array).

hmmlearn/base.py
sagunb: I would think it would be a more compact model if we saved these three variables, rather than the full history of log probabilities. (For my usage case right now, we sometimes need 100k + iterations . And if we do ensemble methods, if each model is storing an array of 100k+ values, that is a huge consumption of memory that is not required). It would be much more efficient (memory wise) to simply store the final logprob, number of iterations performed, and the last delta achieved.

superbobry: Yes, storing only the last `logprob_` and the delta makes sense. 

I do have one last suggestion: would you mind using a shorter name for `n_iterations_performed`, for example `n_iter_performed_`, so that it looks similar to the parameter `n_iter`?

sagunb: Done. Please pull when you have time :)

superbobry: Looks good! 

I think we should also amend `BaseHMM` docstring to reflect the newly added attributes (similar to how it's done for [`sklearn.mixture.GMM`](http://scikit-learn.org/stable/modules/generated/sklearn.mixture.GMM.html#sklearn.mixture.GMM)). Sorry for not mentioning this right away.

sagunb: Done. Please review and merge :)

sagunb:
Added docstring to `BaseHMM` please pull and merge.

superbobry:
Looks good, thanks!

PR #26, opened 2015-04-27
Fix setup script

matthiasplappert:
The package `hmmlearn.util` does not exist anymore, so `setup.py` currently fails. This fixes the problem.


superbobry:
Hm, the `hmmlearn.utils` should still be in the [repo](https://github.com/hmmlearn/hmmlearn/blob/master/hmmlearn/utils.py).

matthiasplappert:
Yes, but it's no longer a dedicated package but rather a module. Try running the current version of setup.py for installation and you'll see.

superbobry:
Yup, sorry, I misunderstood your PR. And thanks for spotting this!

PR #31, opened 2015-06-01
Add a method to check if parameters are identifiable

jaelynwu:
Sometimes EM seems to converge but the parameters are unidentifiable. i.e, two parameters are supposed be distinct but have the same distribution. We need to specify some identification condition(e.g, the distribution must be different) for validation.

To know more about this issue please go to the following links:
http://www.researchgate.net/post/Does_the_EM_algorithm_always_converge
http://en.wikipedia.org/wiki/Identifiability

### DIFF
--- a/hmmlearn/base.py
+++ b/hmmlearn/base.py
@@ -65,6 +65,7 @@ class ConvergenceMonitor(object):
 
         self.history.append(logprob)
         self.iter += 1
+        
 
     @property
     def converged(self):
@@ -405,6 +406,13 @@ class _BaseHMM(BaseEstimator):
                 currstate, random_state=random_state))
 
         return np.array(obs), np.array(hidden_states, dtype=int)
+    
+    def _is_identifiable(self, means, n_components):
+        for i in range(0, n_components):
+            for j in range(0, n_components):
+                if i != j and (means[i] == means[j]).all():
+                    return False
+        return True
 
     def fit(self, obs):
         """Estimate model parameters.
@@ -438,7 +446,10 @@ class _BaseHMM(BaseEstimator):
 
             self.monitor_.report(curr_logprob)
             if self.monitor_.converged:
-                break
+                if not hasattr(self, '_means_'):
+                    break
+                elif hasattr(self, '_means_') and self._is_identifiable(self._means_, self.n_components):
+                    break
 
             self._do_mstep(stats, self.params)
 


hmmlearn/base.py
```diff
@@ -405,6 +406,13 @@ def sample(self, n=1, random_state=None):
                 currstate, random_state=random_state))
 
         return np.array(obs), np.array(hidden_states, dtype=int)
+    
+    def _is_identifiable(self, means, n_components):
```
superbobry: I think this belongs to the `_init` method of each particular subclass, e.g. `GaussianHMM`.

jaelynwu: I am not sure I get what you mean. Indeed not all subclasses need this method, but it is applied in EM  iteration to check if the parameters have converged or converged properly. The argument means is not means_prior in _init but an estimation after e step. The problem is that there is only one generic fit method in which EM steps are conducted and that's why I call  _is_identifiable  in fit method.  I am open to any suggestion though.

superbobry: > I am not sure I get what you mean. 
Sorry, I'll try to explain my point more clearly this time :)

> The problem is that there is only one generic fit method in which EM steps are conducted and that's why I call _is_identifiable in fit method.

I see two ways to overcome this problem:

1. Add `_is_identifiable` property to `_BaseHMM`. The property should be `True` by default. The subclasses are free to override it and check the parameters for identifiability. This in fact is a slight modification of what you're proposing in the PR. The only difference is that `_is_identifiable` is a property, not a method. 
2. Do not change `_BaseHMM` and add diagnostic step to `_m_step` of the relevant models. The method is expected to issue a warning if parameters aren't identifiable.

The former way implies that EM iterates until it finds identifiable parameters. While in theory this might help with pathological cases, I'm not sure it is very useful in practice. Can you give an example of the dataset where EM first converges to equal means and then after a number of iterations uncovers the true means?

P. S. There's also a problem of defining `equals` for real-valued parameters. For some applications `0.0001` is equal to `0.00012` for other not. And since we are operating in floating point we **have** to pick the threshold somehow.

jaelynwu: Thanks for your elaboration. I will experiment with the two options you
mentioned. In fact the parameter identifiability issues has happened to me
a few times in my work, where the iterations stop because the error is
below the threshold but the parameters estimated are useless. Since I
cannot share the raw data,  I need to manipulate it a bit b. Please let me
know how before sending to you.  Please let me know how you prefer to
receive the dataset.

PS, good point about "equals".  I used a threshold myself, but the problem
is the user might have to adjust the threshold actively depending on the
expected distribution(not necessarily a bad thing but can be a hassle).
Still I agree picking a threshold is better option.

jaelynwu: Attached is an example dataset(not sure you will be able to see that
properly, let me know otherwise) with which you would have parameter
identifiability
issue.  Each line is a sequence of observations, and the value of each
observation in a sequence is a 1x1 array. The number of components is 3.
After EM converges, the estimated means and the variance of the 3 states
are as follows:

0th hidden state
('mean = ', array([ 0.]))
('var = ', array([  2.44762728e-07]))
()
1th hidden state
('mean = ', array([ 3.17144563]))
('var = ', array([ 23.84039865]))
()
2th hidden state
('mean = ', array([ 0.]))
('var = ', array([  2.44845276e-07]))
()

As you can see, the distribution of the state 0 and 2 are exactly the same,
meaning they are unidentifiable.

anntzer: I think the floating-point equality problem is a real one here.  (Even though there can be cases of exact equality as demonstrated above, in my experience approximate equality is much more common.)

Issue #33, opened 2015-06-03
Is there support for non-fully connected HMMs?


oxinabox:
As far as I can tell there is no support for nonfully connected HMMs.
Eg Left-Right HMMs,
where each start can only move to states of index greater than, or equal to its own.

Am I correct in saying that they are not supported, for training?
Or have I missed something?
(I thought for a moment, that setting the transition_mat to have -Inf in the disallowed transitions would work, but it does not).

They can be implemented if you don't want to fit them., simply by setting the non-connected state probabilities to zero.
But doing thing and then asking it to `fit` causes an error.

matthiasplappert:
I'm also curious about this. My current approach is to set some entries in the transition matrix to 0. However, they will get replaced with a value very close to zero (e.g. 1e-15) when normalizing. Is this sufficiently small to allow for different topologies?

superbobry:
In theory `hmmlearn` should support non-fully connected HMM. I think the error you're seeing is due to the validation logic. I'll check that this is the case and update the issue accordingly.

oxinabox:
Can a example be added to the documentation for how to train and use them?

cliberatore:
I'm with @superbobry on this one--in my experience with the framework, it's possible to (implicitly) have non-fully-connected HMMs. I've got some data I'm working with which is inherently left-to-right, and whether or not I initialize with a transition matrix which forces left-to-right topology (i.e. eye(n_components)), I'll almost always end up with a transition matrix which reflects this desired structure.

However, as @matthiasplappert noted, you won't get 0 transition probabilities, just extremely low values. In practice, it strikes me that this is just fine--when you have transition probabilities which are more than 10 orders of magnitude below other entries, those transitions never occur in practice.

I would make an argument that, as provided, this module supports non-fully-connected HMMs.

matthiasplappert:
I've actually found that, currently, hmmlearn does not really support non-fully-connected HMMs. This is caused by the normalization of the transition matrix that is applied when training the HMM. Here's a very simple example:

```
import hmmlearn.hmm as hmm
import numpy as np

n_states = 3

# Left-to-right: each state is connected to itself and its direct successor
transmat = np.zeros((n_states, n_states))
for i in range(n_states):
	if i == n_states-1:
		transmat[i, i] = 1.0
	else:
		transmat[i, i] = 0.5
		transmat[i, i+1] = 0.5

# Always start in first state
startprob = np.zeros(n_states)
startprob[0] = 1.0

[...]

PR #38, opened 2015-07-22
Support for non-fully connected topologies

matthiasplappert:
See discussion in https://github.com/hmmlearn/hmmlearn/issues/33

This change still needs tests. However, I won't have time to work on them until August.


superbobry:
Hello, @matthiasplappert! Your solution is quite elegant, but unfortunately it cannot be merged as is due to the API changes in version `0.2.0`. 

Do you think it's a good idea to use the same (n_components, )-shaped mask for both `startprob_` and `transmat_`?

PR #44, opened 2015-08-10
self._covars_ in GaussianHMM is not writeable when settin n_components=1

dhuppenkothen:
For a GaussianHMM with n_components=1 and covariance_type="diag", the last line in GaussianHMM._init throws a ValueError, because self._covars_ is not writeable (flags.["WRITEABLE"]=False). 

This is caused by the use of numpy.diag in sklearn.mixture.distribute_covar_matrix_to_match_covariance_type, which returns a read-only numpy.ndarray. This is intended behaviour for numpy 1.9 (see http://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.diagonal.html). As per the recommendation in the documentation, setting self._covars_ to a copy of itself fixes the issue and makes behaviour consistent across current and future versions of numpy.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -20,3 +20,4 @@ Authors and contributors, in no particular order:
 * csytracy
 * Alexandre Gramfort
 * Sergei Lebedev
+* Daniela Huppenkothen
--- a/hmmlearn/hmm.py
+++ b/hmmlearn/hmm.py
@@ -191,7 +191,9 @@ class GaussianHMM(_BaseHMM):
                 cv.shape = (1, 1)
             self._covars_ = distribute_covar_matrix_to_match_covariance_type(
                 cv, self.covariance_type, self.n_components)
-            self._covars_[self._covars_ == 0] = 1e-5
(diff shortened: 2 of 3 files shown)


superbobry:
Good catch! Can you please add a test (specifically for NumPy >= 1.9) to make sure we don't accidentally remove this in the future?

Also, feel free to add yourself to `AUTHORS.rst`.

dhuppenkothen:
I've added a test, which passes with the new code in numpy 1.9, but fails with the new code commented out. I've also tested the code on the latest development version of numpy (1.11.0.dev). It looks like in numpy  1.11, as it is now, numpy.diag still returns a read-only array, but the use of numpy.tile will make it writable again (unlike in numpy 1.9). This is not what I expected given the documentation in the link posted above, but I don't understand enough of the underlying numpy memory management to make sense of it. 

As it is, the test will pass in the development version of numpy with or without the new code.

superbobry:
Looks good. I'll add a reference to this PR, so that we know where to look once the tests starts failing.

PR #51, opened 2015-09-18
adding cython as requirement in setup.py

stringertheory:


superbobry:
As far as I understand, Cython is not required during runtime, so it doesn't belong to `requires`.

PR #56, opened 2015-10-21
DOC incorrect startprob_ in left-right HMM example

soupault:
Hi!
There seems to be a typo in documentation.


superbobry:
Yep, thanks!

PR #82, opened 2016-01-18
ENH: Use typed memoryviews and remove numpy calls in cythonized _viterbi

cfarrow:
Were using hmmlearn's Viterbi algorithm on large sample sets and wanted to speed it up a bit. On our data (millions of samples) we found a 1000x speedup with these changes. I profiled `TestBaseHMM.test_do_viterbi_pass` in `hmmlearn/tests/test_base.py` and found a 4x speedup in the `_viterbi` function.

The speedups are coming from faster indexing with typed memoryviews and removing calls into python from the inner loops, which would incur a penalty with every iteration. The kernel of the Viterbi implementation is now free of python, so I put it in a `nogil` block.


hmmlearn/_hmmc.pyx
```diff
@@ -1,12 +1,11 @@
 # cython: boundscheck=False, wraparound=False
 
 cimport cython
-cimport numpy as np
 from numpy.math cimport expl, logl, isinf, INFINITY
 
 import numpy as np
 
-ctypedef np.float64_t dtype_t
+ctypedef double dtype_t
```
superbobry: Isn't `double` the same as `np.float64_t`?

hmmlearn/_hmmc.pyx
```diff
@@ -27,13 +26,13 @@ cdef inline dtype_t _logsumexp(dtype_t[:] X) nogil:
...
-        np.ndarray[dtype_t, ndim=2] framelogprob,
-        np.ndarray[dtype_t, ndim=2] fwdlattice):
+        dtype_t[:] log_startprob,
+        dtype_t[:, :] log_transmat,
+        dtype_t[:, :] framelogprob,
+        dtype_t[:, :] fwdlattice):
 
     cdef int t, i, j
-    cdef dtype_t[:] work_buffer = np.zeros(n_components)
+    cdef dtype_t[::1] work_buffer = np.zeros(n_components)
```
superbobry: Can you please explain what does the `::1` slice do in this case?

hmmlearn/_hmmc.pyx
```diff
@@ -92,34 +91,58 @@ def _compute_lneta(int n_samples, int n_components,
...
+        for c1 in range(n_components):
+            viterbi_lattice[0, c1] = log_startprob[c1] + framelogprob[0, c1]
 
-    for t in range(n_samples - 2, -1, -1):
-        max_pos = np.argmax(viterbi_lattice[t, :] \
-                + log_transmat[:, state_sequence[t + 1]])
-        state_sequence[t] = max_pos
+        # Induction
+        for t in range(1, n_samples):
+            for c0 in range(n_components):
```
superbobry: We use `i` and `j` for states throughout the codebase. Sorry for being a bit pedantic :)

superbobry:
This looks very good, thank you. I'll merge this right after the minor issues are resolved.

hmmlearn/_hmmc.pyx
cfarrow: Yes. There was a mix in the code so I standardized on `double`. This was unnecessary and I can revert it if you'd like.

hmmlearn/_hmmc.pyx
cfarrow: No worries. I've changed it.

hmmlearn/_hmmc.pyx
cfarrow: This tells cython that the buffer is contiguous. This allows cython to hardcode the strides, which saves a lookup and results in faster indexing. I should be using the `view.contiguous` specifier and have changed this.

superbobry: Wow, I somehow missed this feature. Looks very useful, thanks.

cfarrow:
Thanks!

PR #97, opened 2016-02-25
Vectorize statistics accumulation

anntzer:
This *greatly* accelerates fitting of models with a tied or full covariance matrix.

### DIFF
--- a/hmmlearn/hmm.py
+++ b/hmmlearn/hmm.py
@@ -229,10 +229,10 @@ class GaussianHMM(_BaseHMM):
             if self.covariance_type in ('spherical', 'diag'):
                 stats['obs**2'] += np.dot(posteriors.T, obs ** 2)
             elif self.covariance_type in ('tied', 'full'):
-                for t, o in enumerate(obs):
-                    obsobsT = np.outer(o, o)
-                    for c in range(self.n_components):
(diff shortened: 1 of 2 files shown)


hmmlearn/hmm.py
```diff
@@ -229,10 +229,10 @@ def _accumulate_sufficient_statistics(self, stats, obs, framelogprob,
...
                 stats['obs**2'] += np.dot(posteriors.T, obs ** 2)
             elif self.covariance_type in ('tied', 'full'):
-                for t, o in enumerate(obs):
-                    obsobsT = np.outer(o, o)
-                    for c in range(self.n_components):
-                        stats['obs*obs.T'][c] += posteriors[t, c] * obsobsT
+                # posteriors.T: (nc, nt)
+                # obs[:, :, None] * obs[:, None, :]: (nt, nf, nf)
+                stats['obs*obs.T'] += np.tensordot(
+                    posteriors.T, obs[:, :, None] * obs[:, None, :], 1)
```
superbobry: This looks great. Do you mind using `np.newaxis` instead of `None` and specifying the name of the `axes` argument?

anntzer: Actually, I just realized this can be written even more succintly using Einstein summation, which (I guess?) also removes the need to allocate a temporary for `obs*obs.T`.

Issue #101, opened 2016-03-13
transmat_  doesn't sum to 1.0


naghmehghanooni:
Hello, 

I am using hmmlearn package version 0.2.1. I want to cluster some 2d sequences. I have a transition matrix which sums to 1.0 (0 for some cells of matrix, but at the end they all sum to 1.0) but when I fit the data I get a value error that the transition matrix doesn't sum to 1.0.

```python
model = hmm.GaussianHMM(n_components=70, covariance_type="full", n_iter=100, params="st", init_params="st")
```

I want to give my initial transition matrix to the model and I want the model to change it (as you can see in the above code). But the model changes it in a way that it is not sum to 1.0 anymore. 
Does anyone know what might be wrong? 
This is the Value Error. The matrix in return is not complete. But you can see it doesn't sum to 1.0 anymore as my initial transition matrix. 

```
File "build/bdist.linux-x86_64/egg/hmmlearn/base.py", line 524, in _check
ValueError: rows of transmat_ must sum to 1.0 (got [ 1.          1.          1.          1.          1.          1.          1.
  1.          1.          1.          1.          1.          1.          1.
  1.          1.          1.          1.          1.          1.          1.
  1.          1.          1.          1.          1.          1.          1.
  1.          1.          1.          1.          1.          1.          1.
  1.          1.          1.          1.          1.          1.          1.
  1.          1.          1.          1.          1.          1.          1.
  1.          1.          1.          1.          1.          1.          1.
  1.          1.          1.          1.          1.          1.          1.
  1.          1.          1.          0.66666667  0.66666667  0.66666667
```

Thanks in advance,

superbobry:
Could you attach the transition matrix you're using?

naghmehghanooni:
transMat = np.zeros((70, 70))
for i, j in itertools.product(range(len(transMat)), range(len(transMat))):
    if i == 0:
        if j == i or j == i+1 or j == i+10 or j == i+1+10:
            transMat[i, j] = 1.0/4.0
    elif i == 9:
        if j == i or j == i-1 or j == i+10 or j == i-1+10:
            transMat[i, j] = 1.0/4.0
    elif i == 60:
        if j == i or j == i+1 or j == i-10 or j == i+1-10:
            transMat[i, j] = 1.0/4.0
    elif i == 69:
        if j == i or j == i-1 or j == i-10 or j == i-1-10:
            transMat[i, j] = 1.0/4.0
[...]

Issue #93, opened 2016-02-20
 probability would approach to 0 after several EM iterations


PacoSpike:
When I used GaussianHMM().fit() to train HMM, there is a RuntimeWarning: divide by zero encountered in log. Then I found  that the start probability would approach to 0 after several EM iterations. 
My question is how to avoid probability approaching to 0 ?

aubreyli:
@PacoSpike , which version of hmmlearn are you using?

PacoSpike:
The latest, 0.2

PacoSpike:
@aubreyli

aubreyli:
@PacoSpike , I previously encountered divide by zero error when I was using 0.1.1.
The problem is gone after I upgrade to latest repo.

So, I think it should be helpful if you place your data onto somewhere and provide a link here.

superbobry:
Zero starting probabilities can happen if you have few training sequences or the most likely starting state for each sequence is the same.

PacoSpike:
@aubreyli This is a sample.
[sample.txt](https://github.com/hmmlearn/hmmlearn/files/139453/sample.txt)

aubreyli:
@PacoSpike , I didn't encounter divide by zero error. please check the attachment.

[sample_log.txt](https://github.com/hmmlearn/hmmlearn/files/139526/sample_log.txt)
![sample](https://cloud.githubusercontent.com/assets/7411812/13201792/f49254a4-d8ba-11e5-9437-f710156f76b6.JPG)

PacoSpike:
@aubreyli Thank you very much. I have examined the sample and  your result is the same as mine. However, the warning is real. So I repeat my data and select 2 new samples. When I use the 2 samples to train HMM
[new_sample_2.txt](https://github.com/hmmlearn/hmmlearn/files/139583/new_sample_2.txt)
[new_sample_1.txt](https://github.com/hmmlearn/hmmlearn/files/139584/new_sample_1.txt)

, I notice the warning occurs randomly.
Thanks again.

aubreyli:
@PacoSpike , I can replicate the warning by new_sample_1.txt now.

I believe the warning occurs randomly, which is because Clustering kmeans algorithm is initialized randomly.

For a temporary workaround and a quick check, can you please do the following change in hmm.py?

Change
`kmeans = cluster.KMeans(n_clusters=self.n_components)`

To
`kmeans = cluster.KMeans(n_clusters=self.n_components, random_state = 1)`

[...]

Issue #76, opened 2015-12-16
Non-ergodic transition matrix leads to NaN


superbobry:
To reproduce just use the example from #33.

PR #105, opened 2016-03-21
Handle ZERO start/transition probabilities.

aubreyli:
See #93 and #101 for details

### DIFF
--- a/hmmlearn/utils.py
+++ b/hmmlearn/utils.py
@@ -60,3 +60,18 @@ def iter_from_X_lengths(X, lengths):
 
         for i in range(len(lengths)):
             yield start[i], end[i]
+
+def log_mask_zero(prob):
+    """Computes the log of input probabilities masking divide by zero in log.
+
+    Notes
+    -----
+    During EM iteration of the model fitting, very small intermediate
+    start/transition probabilities could be normalized to Zero, causes
+    a RuntimeWarning: divide by zero encountered in log. This function
+    masks this unharmful warning.
+    """
+    prob = np.asarray(prob)
+    with np.errstate(divide='ignore'):
+        _log_prob = np.where(prob > 0, np.log(prob), 0)
+    return _log_prob
--- a/hmmlearn/base.py
+++ b/hmmlearn/base.py
@@ -11,7 +11,7 @@ from sklearn.utils import check_array, check_random_state
 from sklearn.utils.validation import check_is_fitted
 
 from . import _hmmc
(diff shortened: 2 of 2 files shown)


hmmlearn/utils.py
```diff
@@ -60,3 +60,9 @@ def iter_from_X_lengths(X, lengths):
 
         for i in range(len(lengths)):
             yield start[i], end[i]
+
+def prob_log(prob):
```
superbobry: We used to have a similar function called [`exp_mask_zero`](https://github.com/hmmlearn/hmmlearn/commit/03dd25107b940542b72513ca45ef57da22aac298), which was removed in 144dc5318b83b2385cb14239ef7b30b8e833f902, because it didn't work for non-ergodic transition matrices (see #76).

aubreyli: a quick check difference is that exp_mask_zero() replaced NaN with Zero while prob_log() replaces -inf with Zero? If this is not the point, I'll take a look at the details of #76.

aubreyli: As I checked the example from #33, I didn't see the patch break non-ergodic transition 
matrices model.
================================
`transmat and startprob BEFORE:
[[ 0.5  0.5  0. ]
 [ 0.   0.5  0.5]
 [ 0.   0.   1. ]]
[ 1.  0.  0.]


transmat and startprob AFTER:
[[ 0.50371  0.49629  0.     ]
 [ 0.       0.00117  0.99883]
 [ 0.       0.       1.     ]]
[ 1.  0.  0.]`

superbobry: Yes, you're right. Ignore my comment above :)

Do you have any ideas for a more descriptive name? I would suggest the *mask* thing, but I'm not sure if it's appropriate.

Could you add a comment (or even a proper docstring) referencing the original issue and explaining the motivation for this function?

One final thought. There is context manager [`np.errstate`](http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.errstate.html#numpy.errstate) for turning warning on and off in a limited scope. You could use it instead of `np.seterr`.

aubreyli: we can pick up log_mask_zero back if you think it's acceptable, :)
let me refine the patch to request another round of comment.

superbobry: Everything looks good. Just squash the commits into a single one using rebase.

superbobry:
Thank you for helping out with the issues and for the patch!

PR #107, opened 2016-04-02
Completely rewrite GMMHMM

yanenkoa:
I rewrote GMMHMM from the ground up without the use of GMM from scikit-learn. It is still work in progress, and I haven't yet wrote any tests for it.

### DIFF
--- a/hmmlearn/hmm.py
+++ b/hmmlearn/hmm.py
@@ -11,6 +11,7 @@ The :mod:`hmmlearn.hmm` module implements hidden Markov models.
 """
 
 import numpy as np
+from scipy.misc import logsumexp
 from sklearn import cluster
 from sklearn.mixture import (
     GMM, sample_gaussian,
@@ -463,14 +464,14 @@ class GMMHMM(_BaseHMM):
         String describing the type of covariance parameters to
         use.  Must be one of
 
-        * 'spherical' --- each state uses a single variance value that
+        * "spherical" --- each state uses a single variance value that
           applies to all features;
-        * 'diag' --- each state uses a diagonal covariance matrix;
-        * 'full' --- each state uses a full (i.e. unrestricted)
+        * "diag" --- each state uses a diagonal covariance matrix;
+        * "full" --- each state uses a full (i.e. unrestricted)
           covariance matrix;
-        * 'tied' --- all states use **the same** full covariance matrix.
+        * "tied" --- all states use **the same** full covariance matrix.
 
-        Defaults to 'diag'.
+        Defaults to "full".
 
     startprob_prior : array, shape (n_components, )
         Initial state occupation prior distribution.
@@ -479,8 +480,8 @@ class GMMHMM(_BaseHMM):
         Matrix of prior transition probabilities between states.
 
     algorithm : string
-        Decoder algorithm. Must be one of 'viterbi' or 'map'.
-        Defaults to 'viterbi'.
+        Decoder algorithm. Must be one of "viterbi" or "map".
+        Defaults to "viterbi".
 
     random_state: RandomState or an int seed
         A random number generator instance.
@@ -536,10 +537,10 @@ class GMMHMM(_BaseHMM):
 
     def __init__(self, n_components=1, n_mix=1,
                  startprob_prior=1.0, transmat_prior=1.0,
-                 algorithm='viterbi', covariance_type='full',
+                 algorithm="viterbi", covariance_type="full",
                  random_state=None, n_iter=10, tol=1e-2,
-                 verbose=False, params='stmcw',
-                 init_params='stmcw'):
+                 verbose=False, params="stmcw",
+                 init_params="stmcw"):
         _BaseHMM.__init__(self, n_components,
                           startprob_prior=startprob_prior,
                           transmat_prior=transmat_prior,
@@ -550,33 +551,31 @@ class GMMHMM(_BaseHMM):
         self.n_mix = n_mix
 
     def _init(self, X, lengths=None):
-        super(GMMHMM, self)._init(X, lengths)
+        super(GMMHMM, self)._init(X, lengths=lengths)
 
-        _, n_features = X.shape
-        self.n_features = n_features
+        _, self.n_features = X.shape
 
         main_kmeans = cluster.KMeans(n_clusters=self.n_components,
                                      random_state=self.random_state)
-        main_kmeans.fit(X)
-        labels = main_kmeans.labels_
+        labels = main_kmeans.fit_predict(X)
         kmeanses = []
-        for i, label in enumerate(np.unique(labels)):
+        for label in range(self.n_components):
             kmeans = cluster.KMeans(n_clusters=self.n_mix,
                                     random_state=self.random_state)
             kmeans.fit(X[np.where(labels == label)])
             kmeanses.append(kmeans)
 
-        if 'w' in self.init_params or not hasattr(self, 'weights_'):
+        if 'w' in self.init_params or not hasattr(self, "weights_"):
             self.weights_ = (np.ones((self.n_components, self.n_mix)) /
                              (np.ones((self.n_components, 1)) * self.n_mix))
 
-        if 'm' in self.init_params or not hasattr(self, 'means_'):
+        if 'm' in self.init_params or not hasattr(self, "means_"):
             self.means_ = np.zeros((self.n_components, self.n_mix,
                                     self.n_features))
             for i, kmeans in enumerate(kmeanses):
                 self.means_[i] = kmeans.cluster_centers_
 
-        if 'c' in self.init_params or not hasattr(self, 'covars_'):
+        if 'c' in self.init_params or not hasattr(self, "covars_"):
             cv = np.cov(X.T)
             if not cv.shape:
                 cv.shape = (1, 1)
@@ -609,7 +608,7 @@ class GMMHMM(_BaseHMM):
         cur_covs = self.covars_[state]
         cur_weights = self.weights_[state]
 
-        i_gauss = np.random.multinomial(1, cur_weights)[0][0]
+        i_gauss = np.random.choice(self.n_mix, p=cur_weights)
         mean = cur_means[i_gauss]
         if self.covariance_type == 'tied':
             cov = cur_covs
@@ -619,21 +618,24 @@ class GMMHMM(_BaseHMM):
         return sample_gaussian(mean, cov, self.covariance_type,
                                random_state=random_state)
 
+    def _compute_log_weighted_gaussian_densities(self, X, i_comp):
+        cur_means = self.means_[i_comp]
+        cur_covs = self.covars_[i_comp]
+        if self.covariance_type == 'spherical':
+            cur_covs = cur_covs[:, np.newaxis]
+        log_cur_weights = np.log(self.weights_[i_comp])
+
+        return log_multivariate_normal_density(
+            X, cur_means, cur_covs, self.covariance_type
+        ) + log_cur_weights
+
     def _compute_log_likelihood(self, X):
         n_samples, _ = X.shape
         res = np.zeros((n_samples, self.n_components))
 
         for i in range(self.n_components):
-            cur_means = self.means_[i]
-            cur_covs = self.covars_[i]
-            if self.covariance_type == 'spherical':
-                cur_covs = cur_covs[:, np.newaxis]
-            cur_weights = self.weights_[i]
-
-            densities = np.exp(log_multivariate_normal_density(
-                X, cur_means, cur_covs, self.covariance_type
-            ))
-            res[:, i] = np.log(np.sum(cur_weights * densities, axis=1))
+            log_denses = self._compute_log_weighted_gaussian_densities(X, i)
+            res[:, i] = logsumexp(log_denses, axis=1)
 
         return res
 
@@ -667,12 +669,8 @@ class GMMHMM(_BaseHMM):
 
         prob_mix = np.zeros((n_samples, self.n_components, self.n_mix))
         for p in range(self.n_components):
-            covar = self.covars_[p]
-            if self.covariance_type == 'spherical':
-                covar = covar[:, np.newaxis]
-            prob_mix[:, p, :] = np.exp(log_multivariate_normal_density(
-                X, self.means_[p], covar, self.covariance_type
-            )) * self.weights_[p] + eps
+            log_denses = self._compute_log_weighted_gaussian_densities(X, p)
+            prob_mix[:, p, :] = np.exp(log_denses) + eps
 
         prob_mix_sum = np.sum(prob_mix, axis=2)
         post_mix = prob_mix / prob_mix_sum[:, :, np.newaxis]
--- a/hmmlearn/hmm.py
+++ b/hmmlearn/hmm.py
@@ -555,18 +555,21 @@ class GMMHMM(_BaseHMM):
         self.covariance_type = covariance_type
         self.min_covar = min_covar
         self.n_mix = n_mix
-        self.weights_prior_alphas = weights_prior
-        self.means_prior = {"mus": means_prior[0], "lambdas": means_prior[1]}
+        self.weights_prior_alphas = np.asarray(weights_prior)
+        self.means_prior = {
+            "mus": np.asarray(means_prior[0]),
+            "lambdas": np.asarray(means_prior[1])
+        }
         self.covars_prior = {}
         if covars_prior is not None:
             if (self.covariance_type == "full" or
                     self.covariance_type == "tied"):
-                self.covars_prior["psis"] = covars_prior[0]
-                self.covars_prior["nus"] = covars_prior[1]
+                self.covars_prior["psis"] = np.asarray(covars_prior[0])
+                self.covars_prior["nus"] = np.asarray(covars_prior[1])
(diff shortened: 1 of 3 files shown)


hmmlearn/hmm.py
```diff
@@ -463,14 +463,14 @@ class GMMHMM(_BaseHMM):
         String describing the type of covariance parameters to
         use.  Must be one of
 
-        * "spherical" --- each state uses a single variance value that
```
superbobry: The rest of the code uses double quotes, please do not change this in the PR.

hmmlearn/hmm.py
```diff
@@ -520,134 +520,221 @@ class GMMHMM(_BaseHMM):
...
+        self.n_mix = n_mix
 
     def _init(self, X, lengths=None):
-        super(GMMHMM, self)._init(X, lengths=lengths)
+        super(GMMHMM, self)._init(X, lengths)
 
-        for g in self.gmms_:
-            g.set_params(init_params=self.init_params, n_iter=0)
-            g.fit(X)
+        _, n_features = X.shape
```
superbobry: Assign to `self.n_features` directly:

```python
_n_samples, self.n_features = X.shape
```

hmmlearn/hmm.py
```diff
@@ -520,134 +520,221 @@ class GMMHMM(_BaseHMM):
...
-                gmm = GMM(n_mix, random_state=self.random_state)
-            else:
-                gmm = GMM(n_mix, covariance_type=covariance_type,
-                        random_state=self.random_state)
-            self.gmms_.append(gmm)
+        self.n_mix = n_mix
 
     def _init(self, X, lengths=None):
-        super(GMMHMM, self)._init(X, lengths=lengths)
+        super(GMMHMM, self)._init(X, lengths)
```
superbobry: Please pass `length` as a keyword argument. It's part of the internal API contract.

hmmlearn/hmm.py
```diff
@@ -520,134 +520,221 @@ class GMMHMM(_BaseHMM):
...
+        self.n_features = n_features
 
-    def _compute_log_likelihood(self, X):
-        return np.array([g.score(X) for g in self.gmms_]).T
+        main_kmeans = cluster.KMeans(n_clusters=self.n_components,
+                                     random_state=self.random_state)
+        main_kmeans.fit(X)
+        labels = main_kmeans.labels_
+        kmeanses = []
+        for i, label in enumerate(np.unique(labels)):
```
superbobry: The number of unique labels is always `n_components`.

hmmlearn/hmm.py
```diff
@@ -520,134 +520,221 @@ class GMMHMM(_BaseHMM):
...
-            g.set_params(init_params=self.init_params, n_iter=0)
-            g.fit(X)
+        _, n_features = X.shape
+        self.n_features = n_features
 
-    def _compute_log_likelihood(self, X):
-        return np.array([g.score(X) for g in self.gmms_]).T
+        main_kmeans = cluster.KMeans(n_clusters=self.n_components,
+                                     random_state=self.random_state)
+        main_kmeans.fit(X)
```
superbobry: I think you can do `fit_predict` here.

hmmlearn/hmm.py
```diff
@@ -520,134 +520,221 @@ class GMMHMM(_BaseHMM):
...
-        return self.gmms_[state].sample(1, random_state=random_state).flatten()
+        # TODO: actually use random_state
+        if random_state is None:
+            random_state = self.random_state
+
+        cur_means = self.means_[state]
+        cur_covs = self.covars_[state]
+        cur_weights = self.weights_[state]
+
+        i_gauss = np.random.multinomial(1, cur_weights)[0][0]
```
superbobry: You want Categorical distribution here. Try [`np.random.choice`](http://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.random.choice.html) with `p` argument.

hmmlearn/hmm.py
```diff
@@ -520,134 +520,221 @@ class GMMHMM(_BaseHMM):
...
+            cur_means = self.means_[i]
+            cur_covs = self.covars_[i]
+            if self.covariance_type == 'spherical':
+                cur_covs = cur_covs[:, np.newaxis]
+            cur_weights = self.weights_[i]
+
+            densities = np.exp(log_multivariate_normal_density(
+                X, cur_means, cur_covs, self.covariance_type
+            ))
+            res[:, i] = np.log(np.sum(cur_weights * densities, axis=1))
```
superbobry: This is usually done using [`logsumexp`](http://docs.scipy.org/doc/scipy-0.16.0/reference/generated/scipy.misc.logsumexp.html) from `scipy.misc`:

```
res[:, i] = logsumexp(log_multivariate_normal_density(...) + cur_weights, axis=1)
```

hmmlearn/hmm.py
```diff
@@ -520,134 +520,221 @@ class GMMHMM(_BaseHMM):
...
+
+    def _compute_log_likelihood(self, X):
+        n_samples, _ = X.shape
+        res = np.zeros((n_samples, self.n_components))
+
+        for i in range(self.n_components):
+            cur_means = self.means_[i]
+            cur_covs = self.covars_[i]
+            if self.covariance_type == 'spherical':
+                cur_covs = cur_covs[:, np.newaxis]
```
superbobry: Why do we need explicit broadcast here? I would expect `self.covars_` to match `self.covariance_type`?

hmmlearn/hmm.py
```diff
@@ -520,134 +520,221 @@ class GMMHMM(_BaseHMM):
...
+        # post_comp = posteriors.reshape((n_samples,))
+
+        eps = 1e-15
+
+        prob_mix = np.zeros((n_samples, self.n_components, self.n_mix))
+        for p in range(self.n_components):
+            covar = self.covars_[p]
+            if self.covariance_type == 'spherical':
+                covar = covar[:, np.newaxis]
+            prob_mix[:, p, :] = np.exp(log_multivariate_normal_density(
```
superbobry: Maybe we should have the mixture density as a separate function since it's computed in both `_compute_log_likelihood` and `_accumulate_sufficient_statistics`.

hmmlearn/hmm.py
yanenkoa: For some reason `sklearn.mixture.log_multivariate_normal_density` expects `covars` parameter to be of shape `(n_mix, n_features)`.

hmmlearn/hmm.py
```diff
@@ -520,134 +525,294 @@ class GMMHMM(_BaseHMM):
...
+        # Checking means' shape
+        self.means_ = np.array(self.means_)
+        if self.means_.shape != (self.n_components, self.n_mix,
+                                 self.n_features):
+            raise ValueError("mixture means must have shape "
+                             "(n_components, n_mix, n_features), "
+                             "actual shape: {}".format(self.means_.shape))
+
+        # Checking covariances' shape
+        self.covars_ = np.array(self.covars_)
```
superbobry: We could follow `scikit-learn` here and extract the lengthy checks into separate functions. What do you think?

hmmlearn/hmm.py
```diff
@@ -520,134 +525,294 @@ class GMMHMM(_BaseHMM):
...
+        }
+        needed_shape = needed_shapes[self.covariance_type]
+        if covars_shape != needed_shape:
+            raise ValueError("{!r} mixture covars must have shape {}, "
+                             "actual shape: {}"
+                             .format(self.covariance_type,
+                                     needed_shape, covars_shape))
+
+        # Checking covariances' mathematical correctness
+        from scipy import linalg
```
superbobry: Why not makes this a toplevel import?

hmmlearn/hmm.py
```diff
@@ -520,134 +525,294 @@ class GMMHMM(_BaseHMM):
...
+        )
+
+        n_samples, _ = X.shape
+
+        stats['n_samples'] = n_samples
+        stats['samples'] = X
+
+        # post_comp = posteriors.reshape((n_samples,))
+
+        eps = 1e-15
```
superbobry: You could use [machine epsilon](http://stackoverflow.com/a/19141711/262432) here.

hmmlearn/tests/test_gmm_hmm_new.py
```diff
@@ -0,0 +1,165 @@
...
+        self.n_features = 2
+
+        low, high = 10, 15
+        (self.covars, self.means,
+         self.startprob, self.transmat, self.weights) = prep_params(
+            self.n_components, self.n_mix, self.n_features,
+            self.covariance_type, low, high, self.prng
+        )
+        
+        self.h = GMMHMM(n_components=self.n_components, n_mix=self.n_mix,
```
superbobry: It's just a matter of taset ofc, but you can get away w/o using a class here, see [`pytest.mark.parametrize`](https://pytest.org/latest/parametrize.html).

hmmlearn/tests/test_gmm_hmm_new.py
```diff
@@ -0,0 +1,165 @@
...
+    (n_features,) = low.shape
+    X = np.zeros((n_samples, n_features))
+    for i in range(n_features):
+        X[:, i] = random_state.uniform(low[i], high[i], n_samples)
+    return X
+
+
+def prep_params(n_comps, n_mix, n_features, covar_type,
+                low, high, random_state):
+    # the idea is to generate ``n_comps`` bounding boxes and then
```
superbobry: Good idea! :)

hmmlearn/tests/test_gmm_hmm_new.py
```diff
@@ -0,0 +1,165 @@
...
+
+    means = np.zeros((n_comps, n_mix, n_features))
+    for i, (left, right) in enumerate(zip(dim_lims, dim_lims[1:])):
+        means[i] = sample_from_parallelepiped(left, right, n_mix,
+                                              random_state)
+
+    startprob = np.zeros(n_comps)
+    startprob[0] = 1
+
+    transmat = random_state.uniform(size=(n_comps, n_comps))
```
superbobry: We have `normalized` in `tests/__init__.py`.

superbobry:
Looks good so far. Here's a checklist of what needs to be done before we can merge this:

- [ ] Add a test for increasing log-likelihood.
- [ ] Augment the implementation with priors for mixture component parameters and mixture weights. The argument names should match the current (broken) implementation.
- [ ] Ensure that the implemented M-step correctly handles multiple sequences. `_BaseHMM` simply sums up the sufficient statistics over all of the sequences.
- [ ] Time and compare estimation speed of the new `GMMHMM` to the current implementation.
- [ ] Include complete derivation of the M-step in RST format.

hmmlearn/hmm.py
yanenkoa: Is there really a point in exporting part of code into a separate function, when we're only going to call this function once? I feel like it's only going to obfuscate the code for someone trying to read it.

hmmlearn/hmm.py
yanenkoa: To be honest, I did it, because similar function in scikit-learn does it. But it makes sense, this module is only used here. But I can do this way or another.

hmmlearn/tests/test_gmm_hmm_new.py
yanenkoa: I did it because previous version of tests for GMMHMM and other tests use this pattern. But if we're going to use `pytest.mark.parametrize` and don't use classes, I don't quite see, how are we going to do setup for each function. I don't think it's possible to parametrize setup functions.

superbobry: Yep, I was planning on switching to `parametrize` at some point. You don't have to do this if you don't feel like it.

>  I don't quite see, how are we going to do setup for each function. I don't think it's possible to parametrize setup functions.

We can make `self.h` and `self.prng` fixtures. The rest of the attributes can be extracted from `self.h`, e.g.

```
self.n_components == self.h.n_components.

File: hmmlearn/hmm.py (hunk 11 above)
Comment (superbobry) [inline hmmlearn/hmm.py]: Okay, let's leave it as is for now.

File: hmmlearn/hmm.py (hunk 10 above)
Comment (superbobry) [inline hmmlearn/hmm.py]: Functions can be used to improve readability. Take a look at the code. It's long and complex, so you've put in a comment communicating the intent to the reader. However, you could've just as well used a function:

```python
validate_covars_shape(...)
validate_covars_consistency(...)

# Or just validate_covars(...)
```

So, instead of obfuscating the code, the function can make it easier for the reader to get the "bigger" picture of what's going on. Assuming the function has aa good name of course :)

yanenkoa: Aren't comments help to understand the bigger picture just as well? Knowing myself, when I see a function which is called `do_something`, I never trust it to actually do that, and end up looking in the source code anyway. I can be doing it all wrong, though.

matklad: If we speak about `comment block of code` vs `extract it as a named function`, named functions are preferable imo. Comments tend to become obsolete much faster then function names, and you usually have more motivation for fixing names than for fixing comments (well, at least for me :) ). And interesting way to bind a comment to a piece of code is to write a descriptive commit message (http://mislav.net/2014/02/hidden-documentation/).

But extracting a functions which will be used only once has other advantages. 

Consider a hundred lines long function. To get an idea of how it works, you need to skim through all this lines. However if the function was decomposed into ten more or less isolated and properly named sub functions, you'll have to read only ten lines of code. To make this more concrete, I find [this](https://github.com/netvl/intellij-rust/blob/4206e9102ca510df023ca830945d1601e3e50d79/src/main/kotlin/org/rust/ide/intentions/AddDerive.kt#L20-L27) a lot more readable than [this](https://github.com/netvl/intellij-rust/blob/52ea71b6cf2e0a87d000158befba18460907cf73/src/main/kotlin/org/rust/ide/intentions/AddDerive.kt#L20-L41).

Another advantage is that by extracting a function you explicitly isolate necessary context and pass it as parameters, which immediately gives you a sense of how data flows through the program. This is also an instance of "minimize visibility" advice, which is especially important in the languages without static scopes  :)

yanenkoa: I would agree if any data was actually flowing in this code (and I did exactly this in other parts of code, where data is actually passed around), but here this is basically a script. No data, we just need to execute these lines of code. If I were to extract the code in separate functions, these functions won't take or return any values. And if, say, you use some IDE, which can fold custom blocks of code, commented block of code would be literally the same as separate function. Even if you don't have such a feature, I don't think scrolling is that big of a task.

> Consider a hundred lines long function. To get an idea of how it works, you need to skim through all this lines.

You really don't, if there are proper comments in place. Sure, comments get outdated, but so do function names. And in the case as simple as this, I don't think keeping comments up to date is any harder than function names. In fact, it can be way simpler, if I'm not using IDE with "Rename function" feature.

And the advantage of not using function in our case is that if I want to really understand what the code does, I don't need to hop around the source file, I can just see it all in one place, where it is actually executed. Also, Python does not have private methods, so the namespace is not getting polluted.

superbobry: @matklad, @yanenkoa thanks for elaborating!

I suggest we stop the discussion at this point and leave the code as is. The issue is really a minor (and admittedly a subjective) one and isn't worth any more of your time.

hmmlearn/hmm.py
```diff
@@ -554,12 +555,31 @@ def __init__(self, n_components=1, n_mix=1,
         self.covariance_type = covariance_type
         self.min_covar = min_covar
         self.n_mix = n_mix
+        self.weights_prior_alphas = np.asarray(weights_prior)
+        self.means_prior = {
+            "mus": np.asarray(means_prior[0]),
```
superbobry: In the original version `means_prior[0]` is `means_prior` and `means_prior[1]` is `means_weights`. I think we should avoid breaking API compatibility here.

hmmlearn/hmm.py
```diff
@@ -554,12 +555,31 @@ def __init__(self, n_components=1, n_mix=1,
         self.covariance_type = covariance_type
         self.min_covar = min_covar
         self.n_mix = n_mix
+        self.weights_prior_alphas = np.asarray(weights_prior)
```
superbobry: We try to follow `sklearn` [API conventions](http://scikit-learn.org/stable/developers/contributing.html#apis-of-scikit-learn-objects) in `hmmlearn`

> [...] every keyword argument accepted by __init__ should correspond to an attribute on the instance.

hmmlearn/hmm.py
```diff
@@ -601,12 +621,116 @@ def _init(self, X, lengths=None):
...
+
+    def _fix_priors_shape(self):
+        # If priors are numbers, this function will make them into a
+        # matrix of proper shape
+        if len(self.weights_prior_alphas.shape) == 0:
+            self.weights_prior_alphas = self.weights_prior_alphas * np.ones((
+                self.n_components, self.n_mix
+            ))
+
+        if len(self.means_prior["mus"].shape) == 0:
```
superbobry: `len(array.shape)` is `array.ndim`.

hmmlearn/hmm.py
yanenkoa: What is `means_weights`?

superbobry:
I've decided to merge this and do all the brushing-up in master. Thank you very much, @yanenkoa!

Issue #138, opened 2016-07-27
MultinomialHMM - how can I set prior emission probabilities?


hshteingart:
The init method of MultinomialHMM can get the prior of start probabilities and transition probabilities but not the emission probabilities. This may cause the association between state index in the model and states of the underlying model to be not consistent. E.g. let's say state 0 has p_emission of [0.1, 0.9] (corresponding to symbol 0 and 1, respectively), whereas state 1 has probability of emission [0.9, 0.1]. We would like to have emissions of 1's to be associated with state 0.

superbobry:
You can set `emissionprob_` manually like this

```python
hmm = MultinomialHMM(..., init_params="st")
hmm.emissionprob_ = [...] 
hmm.fit(...)
```

Note that `init_params` doesn't have `"e"`, which means that emission probabilities must be specified explicitly prior to fitting.

PR #143, opened 2016-08-23
Exposed emissionprob_prior for MultinomialHMM

soupault:
Addresses #138.


hmmlearn/hmm.py
```diff
@@ -320,6 +319,9 @@ class MultinomialHMM(_BaseHMM):
         Parameters of the Dirichlet prior distribution for each row
         of the transition probabilities :attr:`transmat_`.
 
+    emissionprob_prior : array, shape (n_components, n_features), optional
+        Emission probabilities for each symbol and state :attr:`emissionprob_`.
```
superbobry: I think we should mention that the prior is also a Dirichlet distribution.

PR #168, opened 2016-12-28
check probability matrix when instantiating multinomial hmm

mynameisvinn:
verify startprob_, transmats_, and emissionprob_ sum up to 1 when instantiating multinomial hmm!

### DIFF
--- a/hmmlearn/base.py
+++ b/hmmlearn/base.py
@@ -504,6 +504,11 @@ class _BaseHMM(BaseEstimator):
     def _check(self):
         """Validates model parameters prior to fitting.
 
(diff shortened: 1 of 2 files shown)


hmmlearn/base.py
```diff
@@ -504,6 +504,11 @@ def _init(self, X, lengths):
     def _check(self):
         """Validates model parameters prior to fitting.
 
+        Note: Only startprob_ and transmat_ are covered by this method, as
```
superbobry: I understand your confusing here, but I think this comment could be inferred from the fact that `_BaseHMM` is a base class and it only validates the parameters it defines.

PR #203, opened 2017-05-18
Typo/grammar correction

dannaf:
Grammar problem in sentence "The hidden states are not be observed directly."


superbobry:
Thanks, Daniel!

Issue #198, opened 2017-05-10
DeprecationWarning: Function sample_gaussian is deprecated


ErickTalarico:
I am getting the above warning from Ipython, and it's annoying because it repeats far many times at once.

Best regards
Thanks for making your code available

Erick Talarico

alexnanchen:
A similar warning: 

"DeprecationWarning: Function log_multivariate_normal_density is deprecated; The function log_multivariate_normal_density is deprecated in 0.18 and will be removed in 0.20"

Many thanks.

Best regards.

Alexandre Nanchen

divinekage:
How do we augment the usage of GMMHMM to avoid the Warning from coming up?

PR #213, opened 2017-07-15
Remove use of deprecated scikit-learn functions

kwohlfahrt:
The number of deprecation warnings shown was getting a little insane (see issue #198). Touched quite a lot of things trying to get tests passing again, so a rough summary is:

Main changes:
- Use `np.random.multivariate_normal` instead of `sample_gaussian`
  - Factor out `fill_covars` to make this work for both Gausian/GMM HMMs
- Create a local copy of `log_multivariate_normal_density`

Other things touched in the process:
- delete old GMMHMM tests. They don't pass since the GMMHMM rewrite
- Fix a bug (I think?) in generating spherical covars for the test, looked like the shape was wrong?
- Clean up generation of covars in test of Gaussian HMM.

I'm not very familiar with this codebase, so please take a close look!


kwohlfahrt:
Only Python 2.6 is failing on travis, due to `np.broadcast_to` not existing. Numpy 1.10, which has `broadcast_to` should be available for python 2.6, so it must be a problem with conda.

Why not just use pip on travis?

superbobry:
Hi Kai, thanks for your contribution! I am not sure about removing GMMHMM tests in the same PR. I know they've been failing, but I'd prefer to target them separately. Could you rebase?

We are not using pip, because at the time I did the CI infrastructure NumPy/SciPy were not shipping binary wheels and compilation took too much time.

kwohlfahrt:
Sure, should I mark them xfail? I can change Travis to use pip, and that should get CI working again? Can split that up into separate PRs?

superbobry:
Thanks, this looks good!

PR #217, opened 2017-07-27
BUG in ``log_mask_zero``: log(0) = -inf

jexner:
Commit f66a053e7f6b1081f000cd871b73b04ea4a62433 introduced masking of
probabilities that are zero to avoid run-time warning of their logarithm.
It was assumed that the log(x) is 0 for x <= 0. But this corresponds to
x = 1. The right value is -inf for x = 0.

This fix sets log(0) = -inf. Additionally, all values of log(x) for
x < 0 are also set to -inf, which is reasonable considering that x is a
probability and should not be smaller than 0 in the first place.

### DIFF
--- a/hmmlearn/utils.py
+++ b/hmmlearn/utils.py
@@ -76,7 +76,7 @@ def log_mask_zero(a):
     a = np.asarray(a)
     with np.errstate(divide="ignore"):
         a_log = np.log(a)
-        a_log[a <= 0] = 0.0
+        a_log[a <= 0] = -np.inf
         return a_log
 
 
--- a/hmmlearn/utils.py
+++ b/hmmlearn/utils.py
@@ -75,9 +75,7 @@ def log_mask_zero(a):
     """
     a = np.asarray(a)
     with np.errstate(divide="ignore"):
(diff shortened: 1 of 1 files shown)


vrenkens:
You changed the line to

'''
 a_log[a <= 0] = -np.inf
'''
you can also just remove the line entirely. np.log(0) is already computed as -inf. You still want to get an error if you get negative values and adding the line is less efficient.

jexner:
@vrenkens You're right. So then, it boils down to `log_mask_zero = np.log`. But we agree that `a_log[a <= 0] = 0.0` is wrong?

vrenkens:
jep. I just have 
```
a = np.asarray(a)
with np.errstate(divide="ignore"):
    return np.log(a)
```

jexner:
If one actually wants the computation to fail for invalid log arguments, you could add `invalid="raise"` to the `errstate`.

superbobry:
Thanks!

kwohlfahrt:
Would it be possible to get a patch release with this fix included? Since it is rather broken without.

PR #218, opened 2017-08-01
Reuse ConvergenceMonitor instance

dslaw:
Allow use of custom monitors, e.g. converge when a threshold value is reached:

```python
from hmmlearn.base import ConvergenceMonitor
from hmmlearn import hmm

class ThresholdMonitor(ConvergenceMonitor):
      @property
      def converged(self):
          return (self.iter == self.n_iter or
                  self.history[-1] >= self.tol)

ghmm = hmm.GaussianHMM(n_components=2, tol=5, verbose=True)
ghmm.monitor_ = ThresholdMonitor(ghmm.monitor_.tol,
                                 ghmm.monitor_.n_iter,
                                 ghmm.monitor_.verbose)
```

### DIFF
--- a/lib/hmmlearn/base.py
+++ b/lib/hmmlearn/base.py
@@ -61,6 +61,11 @@ class ConvergenceMonitor(object):
         return "{0}({1})".format(
             class_name, _pprint(params, offset=len(class_name)))
 
+    def _reset(self):
+        """Reset the monitor's state."""
+        self.iter = 0
+        self.history.clear()
+
     def report(self, logprob):
         """Reports convergence to :data:`sys.stderr`.
 
@@ -174,6 +179,7 @@ class _BaseHMM(BaseEstimator):
         self.n_iter = n_iter
         self.tol = tol
         self.verbose = verbose
+        self.monitor_ = ConvergenceMonitor(self.tol, self.n_iter, self.verbose)
 
     def score_samples(self, X, lengths=None):
         """Compute the log probability under the model and compute posteriors.
@@ -424,7 +430,7 @@ class _BaseHMM(BaseEstimator):
         self._init(X, lengths=lengths)
         self._check()
 
-        self.monitor_ = ConvergenceMonitor(self.tol, self.n_iter, self.verbose)
+        self.monitor_._reset()
         for iter in range(self.n_iter):
             stats = self._initialize_sufficient_statistics()
             curr_logprob = 0
--- a/lib/hmmlearn/base.py
+++ b/lib/hmmlearn/base.py
@@ -45,6 +45,27 @@ class ConvergenceMonitor(object):
 
     iter : int
         Number of iterations performed while training the model.
+
+    Examples
+    --------
+    Use custom convergence criteria by subclassing ``ConvergenceMonitor``
+    and redefining the ``converged`` method. The resulting subclass can
+    be used by creating an instance and pointing a model's ``monitor_``
+    attribute to it prior to fitting.
+
+    >>> from hmmlearn.base import ConvergenceMonitor
+    >>> from hmmlearn import hmm
+    >>>
+    >>> class ThresholdMonitor(ConvergenceMonitor):
+    ...     @property
+    ...     def converged(self):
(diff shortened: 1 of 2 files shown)


superbobry:
I like the idea. Maybe we can approach this in a bit more structured way and allow the user to pass in a monitor instead of `n_iter`, `tol`?

dslaw:
Pass in an instantiated monitor? This doesn't seem like it's worth the cost of making breaking api changes. What about passing the monitor type in as a string (common with `sklearn`) or monitor instance? i.e.

```python
def __init__(self, ..., n_iter=10, tol=.001, monitor="absolute"):
  if isinstance(monitor, Monitor):
    self.monitor_ = monitor
  else:
    self.monitor_ = _monitors[monitor](n_iter, tol, verbose)
  
  self.n_iter = self.monitor_.n_iter
  ...
```
Then we can retain the default behavior for those who are okay with the current convergence method, allow custom monitors to be passed if they inherit from a base class, and provide some common ones as well. Not the greatest solution, as when a monitor instance is passed in, `n_iter`, etc will be ignored and fitting the model will modify the referenced monitor. The latter can be avoided by making a copy, or just warning about it in the docstring, though.

Do you have a preference? (Or another method.)

anntzer:
Right now I think you can already set the monitor_ attribute to whatever you want, so this PR is more about reusing a monitor instance (by resetting it)?  Is that something really needed (what's wrong with just instantiating a new monitor -- if you already went through the bother of writing a monitor subclass, instantiating monitors is easy in comparison...)?

Also reusing a single monitor instance across HMMs would preclude multithreaded code, not that I know whether hmmlearn currently supports that or not but I'd not want to preclude it without a good reason.

dslaw:
It's not necessary that the monitor instance itself be reused, this just allows for the class of the monitor to be changed. Since initialization is in the `fit` method, there's no opportunity to change the monitor class until the model has already been fit, which isn't useful. There's no reason the actual instance needs to be reused, though - instantiating from a user defined class would achieve the same result. The intent was still to keep one monitor per model.

For context, this just seemed like a non-intrusive way to allow for the convergence check to be swapped out. I vaguely recall trying to monkeypatch `ConvergenceMonitor`, too, but forget why it didn't work out.

anntzer:
Ah, I see.
That seems fine; perhaps not the simplest API but backcompatible so heh.
Can you add documentation for this (perhaps indicating that it's mostly the `converged` property of the Monitor class that is intended to be overridden?).  Without documentation that seems to be the kind of features that will end up having exactly one user.

dslaw:
Is there somewhere in particular the documentation should go? I don't think putting it in the docstring of `_BaseHMM` makes much sense.

anntzer:
A note or an example (per https://numpydoc.readthedocs.io/en/latest/format.html#sections) in the docstring of ConvergenceMonitor seems OK?

anntzer:
Also you'll need to rebase...  I changed the layout of the source tree a bit :)
(Otherwise, looks fine to me.)

lib/hmmlearn/base.py
```diff
@@ -45,6 +45,27 @@ class ConvergenceMonitor(object):
...
+    Use custom convergence criteria by subclassing ``ConvergenceMonitor``
+    and redefining the ``converged`` method. The resulting subclass can
+    be used by creating an instance and pointing a model's ``monitor_``
+    attribute to it prior to fitting.
+
+    >>> from hmmlearn.base import ConvergenceMonitor
+    >>> from hmmlearn import hmm
+    >>>
+    >>> class ThresholdMonitor(ConvergenceMonitor):
+    >>>     @property
```
anntzer: fails the doctests.  I guess the continuation lines need to use "...", not ">>>"?

dslaw: TIL! I've never actually run doctest on the examples.

anntzer: neither do it, but it's not my codebase...

anntzer:
Thanks for the PR, sorry it fell off my radar last week.

PR #237, opened 2017-09-22
add a deprecated sklearn function to utils

ZiqianXie:
add a local copy of distribute_covar_matrix_to_match_covariance_type to utils.py since it is deprecated in sklearn.

### DIFF
--- a/hmmlearn/hmm.py
+++ b/hmmlearn/hmm.py
@@ -13,14 +13,12 @@ The :mod:`hmmlearn.hmm` module implements hidden Markov models.
 import numpy as np
 from scipy.misc import logsumexp
 from sklearn import cluster
-from sklearn.mixture import (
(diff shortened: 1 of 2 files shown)


Kornel:
From what I see, log_multivariate_normal_density is deprecated as well. Maybe it would be worth to copy that too?

On the otherhand, I'm wondering why it's deprecated, must be for a reason. Maybe it's because the class GMM is deprecated itself?

Regardless, the build for python 2.6 is failing. I think that's because scikit-learn does not support python 2.6.

 @superbobry what do you think, how should we proceed? I guess copying the two functions seems fair as a quick fix, and I'd drop support for python 2.6 as well.

superbobry:
@Kornel 👍  both suggestions sound good to me.

anntzer:
closed/reopened to retrigger travis.

PR #241, opened 2017-10-09
TST: fix pytest command on appveyor

stevenjkern:
I was investigating a local test failure on the Windows platform and wanted to see if the failure was expected or not, so I went to a recent CI build and noticed that the test suite is not actually being run on Appveyor. This PR modifies the call to `pytest` to make the tests discoverable, though perhaps supplying the path to the installed package might be preferable; I'm not much of a `pytest` expert.

### DIFF
--- a/continuous_integration/appveyor/requirements.txt
+++ b/continuous_integration/appveyor/requirements.txt
@@ -9,8 +9,8 @@
 # of the rackspace folder instead of trying to install from more recent
 # source tarball published on PyPI
 pytest
-numpy==1.9.3
-scipy==0.16.0
+numpy
+scipy==0.19.0
 scikit-learn
 wheel
 wheelhouse_uploader
--- a/continuous_integration/appveyor/requirements.txt
+++ b/continuous_integration/appveyor/requirements.txt
@@ -9,7 +9,7 @@
 # of the rackspace folder instead of trying to install from more recent
 # source tarball published on PyPI
 pytest
-numpy
+numpy==1.13.0
 scipy==0.19.0
 scikit-learn
 wheel
--- a/hmmlearn/tests/test_gaussian_hmm.py
+++ b/hmmlearn/tests/test_gaussian_hmm.py
@@ -24,7 +24,7 @@ class GaussianHMMTestMixin(object):
                                  (1, n_components))
         self.means = prng.randint(-20, 20, (n_components, n_features))
         self.covars = make_covar_matrix(
-            self.covariance_type, n_components, n_features
+            self.covariance_type, n_components, n_features, random_state=prng
         )
 
     def test_bad_covariance_type(self):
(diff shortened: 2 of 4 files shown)


hmmlearn/tests/test_gaussian_hmm.py
```diff
@@ -123,6 +123,7 @@ def test_fit_zero_variance(self):
         h = hmm.GaussianHMM(3, self.covariance_type)
         h.fit(X)
 
+    @pytest.mark.xfail
```
superbobry: I think it works on Mac/Linux. Could you at least mark it as conditional xfail?

stevenjkern: It definitely fails sporadically on linux, given recent TravisCI builds: https://travis-ci.org/hmmlearn/hmmlearn/jobs/285760409#L588-L641, and I can reproduce it locally on linux by running the test many times. For instance, this loop run on `master` just now yielded 4 failures and 96 successes: `$ for i in {1..100}; do pytest hmmlearn/tests/test_gaussian_hmm.py::TestGaussianHMMWithFullCovars::test_fit_with_priors --disable-warnings; done`. For reference, here is the output: [test_fit_with_priors_x100.txt](https://github.com/hmmlearn/hmmlearn/files/1379548/test_fit_with_priors_x100.txt).

I believe the problem has to do with how the covariance matrix is created using only the global `RandomState` rather than providing it with a `RandomState` that is reproducible here: https://github.com/hmmlearn/hmmlearn/blob/ae1a41e4d03ea61b7a25cba68698e8e2e52880ad/hmmlearn/tests/__init__.py#L10-L23

I was planning on fixing that in a separate PR tonight, but I could include that here if that sounds appropriate to you.

continuous_integration/appveyor/requirements.txt
```diff
@@ -9,8 +9,8 @@
 # of the rackspace folder instead of trying to install from more recent
 # source tarball published on PyPI
 pytest
-numpy==1.9.3
-scipy==0.16.0
+numpy==1.13.0
+scipy==0.19.0
```
stevenjkern: I changed the versions here to more recent wheels because once the Appveyor tests started running, the following error was raised during test collection: `ValueError: numpy.dtype has the wrong size, try recompiling. Expected 52, got 56`. We could also try the `scipy-1.0.0rc1` wheels that are on `pypi` presently with the anticipation that `scipy-1.0.0` is scheduled for release next week, along with official `numpy` Windows wheels that are already in the wild.

hmmlearn/tests/test_gaussian_hmm.py
mcg1969: For the record, I'm seeing failures here with 0.2.0 on mac.

anntzer:
https://github.com/hmmlearn/hmmlearn/pull/304 fixes the Appveyor build.  Do you want to extract the other fixes into a separate PR?

anntzer:
I think this is now entirely covered by other PRs, closing, but feel free to ping if I missed anything.
Thanks for the PR.

stevenjkern:
I want going to have time to address this until tonight, so thanks for picking it up. 👍🏻👍🏻

PR #254, opened 2017-11-16
Parallelize forward and backward pass.

karolciba:
Attempt to speed up model learning by introduction of parallelization over components in HMM network in Baum-Welch learning.


karolciba:
It fails for windows platform and unfortunately I do not have access to windows build environment. I've added customization for platform specific flags to enable OpenMP. However it seems that that AppVeyor compilers don't have support for OpenMP.
Do you have any suggestions how to solve that? Is there possibility to enable OpenMP in AppVeyor?

Otherwise I plan to enable OpenMP only for Linux platforms, but then Windows won't benefit of parallelization speed up.

Issue #70, opened 2015-11-29
simple multinomial example


ambushed:
Hi there!

Using the latest master of hmmlearn, I tried running a simple MultinomialHMM example (code below) that results in the following error:

> >  File "build/bdist.macosx-10.5-x86_64/egg/hmmlearn/base.py", line 307, in decode
> > ValueError: could not broadcast input array from shape (6) into shape (1)

Could you please tell me what i am doing wrong?  My expectation is that applying Viterbi should give me the most probable hidden sequence.  However passing a list of observation doesn't work unlike passing a single value which does.

Thanks!

Vlad

``` python
from __future__ import division
import numpy as np
from hmmlearn import hmm

states = ["Rainy", "Sunny"]
n_states = len(states)

observations = ["walk", "shop", "clean"]
n_observations = len(observations)

start_probability = np.array([0.6, 0.4])

transition_probability = np.array([
  [0.7, 0.3],
  [0.4, 0.6]
])

emission_probability = np.array([
  [0.1, 0.4, 0.5],
  [0.6, 0.3, 0.1]
])

model = hmm.MultinomialHMM(n_components=n_states)
model.startprob=start_probability
model.transmat=transition_probability
model.emissionprob=emission_probability

# predict a sequence of hidden states based on visible states
bob_says = [0, 2, 1, 1, 2, 0]
model = model.fit(bob_says)
logprob, alice_hears = model.decode(bob_says, algorithm="viterbi")
print "Bob says:", ", ".join(map(lambda x: observations[x], bob_says))
print "Alice hears:", ", ".join(map(lambda x: states[x], alice_hears))
```

soupault:
Hi!
You need to pass both features (for fitting) and observations (for decode) in 2D numpy-arrays, where the observation index is used as a first dimension one.

Here is the change that fixes your code:

    bob_says = np.array([[0, 2, 1, 1, 2, 0]]).T

Output:

    Bob says: walk, clean, shop, shop, clean, walk
    Alice hears: Rainy, Sunny, Sunny, Sunny, Sunny, Rainy

ambushed:
Thanks a lot!  It helped!

soupault:
Meaningful example, yes. There are already some of them added.
Please, check actual documentation beforce you proceed - https://hmmlearn.readthedocs.org/en/latest/index.html .

If issue is resolved, could you, please, close it?

piyushchauhan2011:
[...]

PR #261, opened 2017-12-04
Create simple HMM example using Python 3+

luciansr:
Adapted from http://sujitpal.blogspot.com.br/2013/03/the-wikipedia-bob-alice-hmm-example.html and https://github.com/hmmlearn/hmmlearn/issues/70 and made it work on latest versions of dependencies (Python3+, for instance)


luciansr:
It compiles and executes but the results don't seem to be correct.. sorry

PR #302, opened 2018-09-27
Fixes

anntzer:
Fixes following API changes and deprecations in latest python, scipy and scikit-learn.


anntzer:
Self-merging to allow pointing people towards using master in case anyone wants to test it before the next release.  Maintainers should feel free to comment more on the PR if things need to be changed.

PR #350, opened 2019-06-24
Mod 1/3 fundamental problem summaries

cstarkjp:
Fix incorrect description of likelihood, which is either the likelihood of the *model* (given the data) or the probability of the data (given the model)


anntzer:
thanks!

PR #354, opened 2019-09-11
Bugfix : negative self-covariances

gasse:
Happens to me occasionally that self-covariances get negative values during training, resulting in nan objective. This ad-hoc patch seems to fix the (numerical ?) issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,5 @@ dist/
 htmlcov/
 oprofile_data/
 .coverage
+.coverage.*
 .gdb_history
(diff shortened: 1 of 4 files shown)


anntzer:
Alternatively, can you at least provide an example dataset to repro the issue.

gasse:
Sorry, the issue was happening in the middle of a script, were data was generated on the fly. I will try record a dataset to replicate. My best bet is that this was occurring when the data consisted in very large values, so probably a numerical issue leading to negative self-covariances.

anntzer:
Closing as I believe https://github.com/hmmlearn/hmmlearn/pull/354#discussion_r323618692 is a typo.  Please feel free to reping if you have a dataset that reproes the problem.
OTOH if there's numerical issues it may be worth investigating more accurate variance computation methods, e.g. https://www.johndcook.com/blog/standard_deviation/.

Issue #285, opened 2018-06-11
ValueError: component 0 of 'full' covars must be symmetric, positive-definite


VertexC:
Hi, I have looked into the previous issue on that but none of them solves my problem. 
I am using GuassianHMM but It doesn't seem to train the model properly.
```python
Traceback (most recent call last):
  File "src/playground/HMM_hmmlearn.py", line 49, in <module>
    prob, state_seq = model.decode(x_m.T)
  File "/Users/bowenchen/State_Space_Model_on_Gene/ssm-env/lib/python3.6/site-packages/hmmlearn/base.py", line 294, in decode
    self._check()
  File "/Users/bowenchen/State_Space_Model_on_Gene/ssm-env/lib/python3.6/site-packages/hmmlearn/hmm.py", line 165, in _check
    super(GaussianHMM, self)._check()
  File "/Users/bowenchen/State_Space_Model_on_Gene/ssm-env/lib/python3.6/site-packages/hmmlearn/base.py", line 524, in _check
    .format(self.transmat_.sum(axis=1)))
ValueError: rows of transmat_ must sum to 1.0 (got [1. 0. 1. 1.])

def initialize():
    g = 0
    while g < G:
        x_m[:, g] = abs(
            np.matrix(np.random.multivariate_normal((theta_m.T * y_m[:, g]).flatten().tolist()[0], np.eye(E)))).T
        g += 1

if __name__ == '__main__':
    initialize()
    model = hmm.GaussianHMM(n_components=K, covariance_type="full",n_iter=30)
    iterate_times = 5
    final_score = -math.inf
    scores = []
    print(x_m)
    while iterate_times > 0:
        model.fit(x_m.T)
        print(model.covars_)
        prob, state_seq = model.decode(x_m.T)
        scores.append(model.score(x_m.T))
        iterate_times -= 1
    print(scores)
```
And there is the print out variable
```
x_m
[[1.68185378e-01 1.02232212e+00 2.68666556e-01 8.49854862e-01
  4.19318133e-01]
 [1.04003708e-01 8.84099494e-01 1.64639541e+00 1.61692603e+00
  1.79678081e+00]
 [5.78889521e-01 6.46720353e-01 1.51399590e+00 1.87850812e-03
  3.36270387e+00]]

covars_
[[[ 0.01243624 -0.02659715  0.03280353]
  [-0.02659715  0.13925868 -0.1131393 ]
  [ 0.03280353 -0.1131393   0.10895525]]

[...]

PR #361, opened 2019-11-09
bugfix for #285 value error; covariances are not positive definite.

Noki24:
Overrided fit() and confirmed covariances are positive definite in the last.


anntzer:
Please explain why you think this is the correct fix; AFAICT the problem is actually as explained in https://github.com/hmmlearn/hmmlearn/issues/285#issuecomment-552104932.

anntzer:
Closing for lack of activity.

Issue #382, opened 2020-05-01
GMMHMM ignores 'params' argument 


bsash:
## I am reporting a bug

GMMHMM fit method is updating even those parameters that it was told not to update through the *params* argument when initializing the object.

In this below example I am setting ```params='st'```, so the ```fit``` method is supposed to not update the *means, weights, covars*. But it is still updating them.

```
from hmmlearn import hmm
import numpy as np

# Parameters of the GMMHMM
n_comps = 2
n_mix = 2
n_features = 1

# Specify the desired parameters for GMM
desired_weights = np.array([[0.6, 0.4], [0.1, 0.9]])
desired_means = np.array([[10, 20], [30, 40]])
desired_covars = np.array([[2, 3], [4, 5]])

# Init the GMM parameters in correct size
weights_prior = np.zeros((n_comps, n_mix))
means_prior = np.zeros((n_comps, n_mix, n_features))
covars_prior = np.zeros((n_comps, n_mix))

# Put the desired parameters into the priors
weights_prior = desired_weights
means_prior[:,:,0] = desired_means
covars_prior = desired_covars

# Initialize the GMMHMM and print the parameters
gmm_hmm = hmm.GMMHMM(n_components=n_comps, n_mix=n_mix, covariance_type='spherical', params='st', init_params='st')
gmm_hmm.means_ = means_prior
gmm_hmm.weights_ = weights_prior
gmm_hmm.covars_ = covars_prior
print('Before fitting...')
print('means')
print(gmm_hmm.means_)
print('weights')
print(gmm_hmm.weights_)
print('covars')
print(gmm_hmm.covars_)

# Generate data to fit
data_to_fit = np.zeros(1000)
data_to_fit[:300] = 5 + np.random.randn(300)*2
data_to_fit[300:500] = 15 + np.random.randn(200)*3
data_to_fit[500:550] = 25 + np.random.randn(50)*4
data_to_fit[550:] = 50 + np.random.randn(450)*5
data_to_fit = data_to_fit.reshape(-1,1)

# Fit the data
gmm_hmm.fit(data_to_fit)
print('After fitting...')
print('means')
print(gmm_hmm.means_)
print('weights')
print(gmm_hmm.weights_)
print('covars')
print(gmm_hmm.covars_)
```
Output
```
Before fitting...
means
[[[10.]
  [20.]]

 [[30.]
  [40.]]]
weights
[[0.6 0.4]
 [0.1 0.9]]
covars
[[2 3]
[...]

PR #383, opened 2020-05-01
Bug fix:Check params before maximizing parameter

bsash:
Fix for Issue #382

In ```GMMHMM._do_mstep```, the current code is not checking if the parameters to be maximized is present in ```self.params``` or not. As a result the ```params``` argument is not working as intended.

Fixed it by adding a condition before the maximization to check that the parameter ('w', 'm', 'c') to be maximized is present in ```self.params```.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,28 +13,28 @@ strategy:
       vmImage: 'ubuntu-16.04'
       pythonVersion: '3.8'
     macOS_py35:
-      vmImage: 'macOS-10.13'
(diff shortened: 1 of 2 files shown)


anntzer:
looks good to me; please also add an entry to CHANGES.rst.

anntzer:
Thanks, I merged your PR manually in https://github.com/hmmlearn/hmmlearn/commit/ef6fd13f86dea6a6bdcc62fcc2e6cd4a5b6f8ad9.

Issue #410, opened 2021-04-23
GMMHMM models training not converging (?)


giorgiolbt:
Hi all,
I am having a problem when trying to fit multiple GMMHMM models to solve a classification problem of emotions recognition from speech samples. 
Basically, the models often don't converge: even if the monitor reports 'True' if printed, I can see in the history that the likelihood is not strictly increasing. Actually, it decreases at some point and the training procedure stops.

Here, I report only the procedure for training one of the models (I should have seven, each one trained with a different training set). The data loaded are attached: [data_training.npy.zip](https://github.com/hmmlearn/hmmlearn/files/6365097/data_training.npy.zip)

    from hmmlearn import hmm
    import numpy as np 
    data = np.load('data_training.npy', allow_pickle=True)

    hmm = hmm.GMMHMM(n_components=2, n_mix=2,n_iter=1000, covariance_type="diag", verbose=True ) 

    X_sequence_concat = np.concatenate(data) 
    lengths = []
    for el in data:
        lengths.append(len(el))

    hmm.fit(X_sequence_concat, np.array(lengths))   
    print("Is the HMM training converged? " + str(hmm.monitor_.converged))

In my actual implementation I have to do this for seven different models and sometimes I get this problem and sometimes I don't, as you can see from the results reported below:

<img width="424" alt="Schermata 2021-04-23 alle 13 14 43" src="https://user-images.githubusercontent.com/37977090/115863608-0096bc80-a436-11eb-8f96-89603d4c82b3.png">

Can you please help me? I'm really struggling with this and I can't find a possible cause of the problem.

Thanks in advance!

anntzer:
Per the issue template:

- Provide a short, self-contained, correct example (http://sscce.org/).
  Maintainers must be able to reproduce the issue **by copy-pasting your
  example in a new, empty file/console/notebook**.  Avoid including data files
  as much as possible; if they are absolutely necessary, they must be in csv or
  npy format, and code for loading the file must be included in the example.

Issues not reproducible with the latest release of hmmlearn or without a short,
self-contained, correct example **will be closed**.

giorgiolbt:
Thanks, I updated the issue in order to respect the template.

anntzer:
Thanks.  I can reproduce the failure.
[...]

PR #417, opened 2021-07-04
GMM-HMM - experimental fix for multi-sequence fit

fcostin:
GMM-HMM - experimental fix for multi-sequence fit
    
Prior implementation of GMMHMM
_accumulate_sufficient_statistics had a limitation
where it failed to accumulate information from
multiple sequences during failing.
    
This caused fitting GMMHMM to multi-sequence data
not to function correctly, as it would not consider
contributions from all sequences when updating GMM
parameters during EM iterations.
    
Improves behaviour for issue 410 on github.
    
Ref: https://github.com/hmmlearn/hmmlearn/issues/410


lib/hmmlearn/tests/test_gmm_hmm_multisequence.py
```diff
@@ -0,0 +1,270 @@
...
+
+    permutations = make_permutations(sequence_data)
+    converged = np.zeros(shape=(len(permutations), ), dtype=bool)
+    scores = np.zeros(shape=(len(permutations), ), dtype=np.float64)
+    for i, p in enumerate(permutations):
+        sequences = sequence_data[p]
+        X = np.concatenate(sequences)
+        lengths = np.asarray([len(seq) for seq in sequences])
+        converged[i], scores[i] = attempt_fit(X, lengths)
+    assert np.all(converged) and np.allclose(scores, np.mean(scores), rtol=5e-03)
```
fcostin, line 269: i have no theoretical basis for choosing an rtol of 5e-3

lib/hmmlearn/tests/test_gmm_hmm_multisequence.py
```diff
@@ -0,0 +1,270 @@
...
+
+        return self.history[-1] - self.history[-2] < self.tol
+
+
+def attempt_fit(X, lengths):
+    model = hmm.GMMHMM(
+        n_components=2,
+        n_mix=2,
+        n_iter=100,
+        covariance_type="diag",
```
fcostin, line 202: maybe should parametrise to cover all covariance_type supported by GMMHMM

lib/hmmlearn/tests/test_gmm_hmm_multisequence.py
```diff
@@ -0,0 +1,270 @@
...
+#
+# This data is a reduced subset of data_training.npy
+# from issue https://github.com/hmmlearn/hmmlearn/issues/410
+# illustrating GMMHMM fit diverging during EM iterations.
+#
+# Transformations to reduce data volume:
+# - keep only first 3 of many sequences, discard rest
+# - keep only first 50 of many samples per sequence, discard rest
+# - keep only first 4 of 17 features per sample, discard rest
+EXAMPLE_SEQUENCES_ISSUE_410_PRUNED = np.array([
```
fcostin, line 19: this input data is not a minimal example, but is a lot smaller than the data attached to the original issue.

i don't understand GMMHMM well enough to know how to construct synthetic input that might be appropriate for testing this

lib/hmmlearn/tests/test_gmm_hmm_multisequence.py
fcostin: this is better as comment in test, not in the pr.

anntzer:
This basically looks fine to me, did you plan to make other changes? (per https://github.com/hmmlearn/hmmlearn/issues/410#issuecomment-873493952)

fcostin:
Thank you for the review, I don't plan to make further substantial changes

> Consider fixing style per flake8, though (no need to overdo it, but at least you can try to linewrap at 79 characters).

Will do.

> Also consider squashing the commits (unless you consider the commit history important).

Will do.

Once I have fixed the style I will compress the history so there is a single test commit followed by single patch commit that makes the test pass

fcostin:
@anntzer requested style changes made & commit history cleaned up.

Please take another look.


When I run the test suite locally with only the new first test commit but not the second GMMHMM multi sequence patch commit, I see:

```
FAILED lib/hmmlearn/tests/test_gmm_hmm_multisequence.py::test_gmmhmm_multi_sequence_fit_invariant_to_sequence_ordering[diag]
FAILED lib/hmmlearn/tests/test_gmm_hmm_multisequence.py::test_gmmhmm_multi_sequence_fit_invariant_to_sequence_ordering[spherical]
FAILED lib/hmmlearn/tests/test_gmm_hmm_multisequence.py::test_gmmhmm_multi_sequence_fit_invariant_to_sequence_ordering[tied]
FAILED lib/hmmlearn/tests/test_gmm_hmm_multisequence.py::test_gmmhmm_multi_sequence_fit_invariant_to_sequence_ordering[full]
```

i.e. the new test is able to detect the problem for each of the four covariance types, and then passes after the patch is applied.

lib/hmmlearn/tests/test_gmm_hmm_multisequence.py
```diff
@@ -0,0 +1,290 @@
...
+    @property
+    def converged(self):
+        if len(self.history) < 2:
+            return False
+
+        if np.isnan(self.history[-1]):
+            raise FailedToConverge("logprob must not be nan")
+
+        if self.history[-1] < self.history[-2] - self.tol:
+            raise FailedToConverge("logprob must be non-decreasing")
```
anntzer, line 205: I would just return False in these two cases for the purpose of making test implementation simpler (or you can use `assert`s and rely on pytest converting assert failures to test failures).

lib/hmmlearn/tests/test_gmm_hmm_multisequence.py
```diff
@@ -0,0 +1,290 @@
...
+    destroy the the state transition structure in the input data.
+
+    See issue 410 on github:
+    https://github.com/hmmlearn/hmmlearn/issues/410
+    """
+    sequence_data = EXAMPLE_SEQUENCES_ISSUE_410_PRUNED
+
+    permutations = make_permutations(sequence_data)
+    converged = np.zeros(shape=(len(permutations),), dtype=bool)
+    scores = np.zeros(shape=(len(permutations),), dtype=np.float64)
```
anntzer, line 271: converged and scores can just be lists that you append to.

anntzer:
Just a couple of nitpicks left.

lib/hmmlearn/tests/test_gmm_hmm_multisequence.py
fcostin: When the monitor `converged` property returns `False` the calling `fit` method of `_BaseHMM` interprets that as "we haven't converged yet, let's keep iterating" until max iters are exhausted. The monitor doesn't seem to have a defined way to signal to `_BaseHMM` that something has gone badly wrong and the fit method abort. 

If we sense nans or decreasing logprod during iterations it seems more helpful (to maintainer trying to understand cause of test failure from symptoms) to immediately abort at first sign something is wrong instead of keeping on running.  So of the two alternatives you propose, using assertions seems superior to me than returning False, as that will abort the fit.

fcostin: switching to use asserts here produces arguably superior logs when the test fails: instead of "assert all [ false false false false ]" it fails quicker and pytest dumps the entire strict monitor state that shows the logprob history values decreasing.

i've removed the exception handling. remains of `attempt_fit` method were trivial, inlined that into sole call site in the test body.

fcostin:
@anntzer - thank you for the prompt review pass, i have addressed your comments, the test is a little leaner and meaner now.

if there are any remaining nits in the code that you are willing to fix up yourself, you are not stepping on my toes if you prefer to just fix them if it is a more expedient way to close off the pr

anntzer:
Thanks, this looks great to me!

PR #426, opened 2021-11-05
Add support for HMMs to be computed using the Scaling algorithm

blckmaxima:
Training HMMs using the scaling version of the forward/backward algorithm is much more efficient - A benchmarking script is included to demonstrate this.  However, I have left the default training algorithm as "log".


examples/benchmark.py
```diff
@@ -0,0 +1,318 @@
...
+import argparse
+import collections
+import logging
+import time
+
+import hmmlearn.hmm
+
+import numpy as np
+
+import pandas as pd
```
anntzer, line 14: There's currently no dependency on pandas and I would rather keep things like that.

lib/hmmlearn/_hmmc.pyx
```diff
@@ -94,6 +188,8 @@ def _compute_log_xi_sum(int n_samples, int n_components,
                         dtype_t[:, :] framelogprob,
                         dtype_t[:, :] log_xi_sum):
 
+    """
+    """
```
anntzer, line 192: no need for these

lib/hmmlearn/hmm.py
```diff
@@ -232,11 +240,8 @@ def _initialize_sufficient_statistics(self):
                                            self.n_features))
         return stats
 
-    def _accumulate_sufficient_statistics(self, stats, obs, framelogprob,
-                                          posteriors, fwdlattice, bwdlattice):
-        super()._accumulate_sufficient_statistics(
-            stats, obs, framelogprob, posteriors, fwdlattice, bwdlattice)
```
anntzer, line 238: what happened here?

lib/hmmlearn/hmm.py
```diff
@@ -903,12 +916,8 @@ def _initialize_sufficient_statistics(self):
         stats['centered'] = []
         return stats
 
-    def _accumulate_sufficient_statistics(self, stats, X, framelogprob,
-                                          post_comp, fwdlattice, bwdlattice):
```
anntzer, line 907: ditto

lib/hmmlearn/_hmmc.pyx
```diff
@@ -1,5 +1,7 @@
 # cython: language_level=3, boundscheck=False, wraparound=False
-
+"""
+Good Readable Reference: https://www.cs.sjsu.edu/~stamp/RUA/HMM.pdf
```
anntzer, line 3: I would rather stick to Rabiner1989, which is well established and I think covers the same algorithm (equations 90+).

anntzer:
Thanks for the contribution.  I agree that the scaling version is much more standard and should be preferred.  However the patch is really large; can you try to undo any purely stylistic changes so that I can concentrate on the implementation issues?  You can always do a followup PR for style concerns.

examples/benchmark.py
blckmaxima: Great Point, I will remove this dependency.

lib/hmmlearn/_hmmc.pyx
blckmaxima: Absolutely.

lib/hmmlearn/_hmmc.pyx
blckmaxima: Updated

lib/hmmlearn/hmm.py
blckmaxima: The update step for a single iteration is now done in two separate steps:
1. Update the transition / start probabilities using a separate method for scaling / logarithms)
2. Update the emission probabilities using a common method for scaling / logarithms
This replaces the single method, that used inheritance / `super()` to update all probabilities in one go.

blckmaxima:
Thank you for reviewing.  Please correct me, but most of the stylistic differences may be in the tests, where I changed all tests to loop over "implementations".  Would you like those changes put in a separate MR ?

Or:
1. one MR that adds the cython methods
2. one MR that adds scaling support to HMMs
3. one MR that adds the changes for tests

Let me know your thoughts.

lib/hmmlearn/_hmmc.pyx
anntzer: Also, these should go to the tutorial, which already has some bibliography entries.

examples/benchmark.py
```diff
@@ -0,0 +1,311 @@
...
+            .5
+        ]).reshape(4, 1, 1,)
+
+        sampler = hmmlearn.hmm.GaussianHMM(
+            n_components=4,
+            covariance_type="full",
+            init_params="",
+            verbose=self.verbose
+        )
+        sampler.startprob_ = pi
```
anntzer, line 104: I would directly define the arrays there, instead of doing it in two steps (same comment applies throughout).

examples/benchmark.py
```diff
@@ -0,0 +1,311 @@
...
+            [.1, .5, .1, .3],
+            [.1, .2, .4, .3],
+            [0, .5, .5, .0],
+        ])
+        sampler = hmmlearn.hmm.MultinomialHMM(n_components=3)
+        sampler.startprob_ = startprob
+        sampler.transmat_ = transmat
+        sampler.emissionprob_ = emissions
+
+        sequences, states = sampler.sample(50_000)
```
anntzer, line 145: we still support py35, which doesn't support underscores in literals.

lib/hmmlearn/_hmmc.pyx
```diff
@@ -1,5 +1,4 @@
 # cython: language_level=3, boundscheck=False, wraparound=False
-
```
anntzer, line 2: please undo all these blank line changes.

lib/hmmlearn/base.py
```diff
@@ -139,8 +139,9 @@ class _BaseHMM(BaseEstimator):
 
     def __init__(self, n_components=1,
                  startprob_prior=1.0, transmat_prior=1.0,
-                 algorithm="viterbi", random_state=None,
-                 n_iter=10, tol=1e-2, verbose=False,
+                 algorithm="viterbi", implementation="log",
```
anntzer, line 142: Technically, this is a backwards incompatible change (in case someone passed all the arguments positionally).  I agree it is most unlikely to affect anyone, but if we're going to break this, we may as well make all args (say, starting from startprob_prior) keyword-only, and add a note in the changelog.  (The whole things needs a changelog entry as well, anyways.)

lib/hmmlearn/base.py
```diff
@@ -262,10 +265,10 @@ def _score(self, X, lengths=None, *, compute_posteriors):
         sub_posteriors = [np.empty((0, self.n_components))]
         for sub_X in _utils.split_X_lengths(X, lengths):
             framelogprob = self._compute_log_likelihood(sub_X)
-            logprobij, fwdlattice = self._do_forward_pass(framelogprob)
+            logprobij, fwdlattice = self._do_forward_log_pass(framelogprob)
```
anntzer, line 268: why does _score not use the scaling impl. if requested by the user?

lib/hmmlearn/base.py
```diff
@@ -469,32 +471,87 @@ def fit(self, X, lengths=None):
...
+        fwdlattice = np.zeros((n_samples, n_components))
+        scaling_factors = np.zeros(n_samples)
+        success = _hmmc._forward_scaling(n_samples, n_components,
+                                       self.startprob_,
+                                       self.transmat_,
+                                       frameprob,
+                                       fwdlattice,
+                                       scaling_factors)
+        if not success:
+            raise ValueError("Error while scaling, try the log pass")
```
anntzer, line 518: Perhaps make the message more explicit ("Consider using implementation='log' instead").

lib/hmmlearn/base.py
```diff
@@ -469,32 +471,87 @@ def fit(self, X, lengths=None):
...
         return self
 
+    def _fit_scaling(self, X, stats):
+        frameprob = self._compute_likelihood(X)
+        logprob, fwdlattice, scaling_factors = self._do_forward_scaling_pass(frameprob)
+        bwdlattice = self._do_backward_scaling_pass(frameprob, scaling_factors)
+        posteriors = self._compute_posteriors_scaling(fwdlattice, bwdlattice)
+        self._accumulate_sufficient_statistics_scaling(
+            stats, X, frameprob, posteriors, fwdlattice,
+        bwdlattice)
```
anntzer, line 481: please fix alignment

lib/hmmlearn/base.py
```diff
@@ -641,6 +714,49 @@ def _initialize_sufficient_statistics(self):
                  'trans': np.zeros((self.n_components, self.n_components))}
         return stats
 
+    def _accumulate_sufficient_statistics_scaling(self, stats, X, frameprob,
```
anntzer, line 717: Instead of introducing a new function, can you not add a switch into the original _accumulate_sufficient_statistics, doing what's required depending on the value of `self.implementation`?

lib/hmmlearn/hmm.py
anntzer: Ah, I see.  See comment above for perhaps re-merging them together and adding an internal `if self.implementation == "scaling": ... elif == "log": ...`

lib/hmmlearn/tests/test_base.py
```diff
@@ -60,20 +59,36 @@ def _compute_log_likelihood(self, X):
...
+                                     [0.9, 0.2],
+                                     [0.1, 0.8],
+                                     [0.9, 0.2],
+                                     [0.9, 0.2]])
+        self.framelogprob = np.log(self.frameprob)
 
         h = StubHMM(2)
-        h.transmat_ = [[0.7, 0.3], [0.3, 0.7]]
-        h.startprob_ = [0.5, 0.5]
+        h.transmat_ = np.asarray([[0.7, 0.3], [0.3, 0.7]])
```
anntzer, line 70: Does it not work anymore to set these as lists?  If so perhaps the implementation should be adjusted to support such inputs.

anntzer:
I understand better the changes now.  There's still quite a few things that need fixing (and I need to check the equations), but probably we should be able to get this in in one go.

examples/benchmark.py
anntzer: edit: actually I'll need to drop py3.5 because there's no manylinux image that supports both py3.5 and py3.10, so no worries there.

blckmaxima:
Thank you for the detailed feedback, it is much appreciated.  I will address these over the next few days.

lib/hmmlearn/_hmmc.pyx
```diff
@@ -64,12 +67,78 @@ def _forward(int n_samples, int n_components,
...
+                     dtype_t[:, :] alpha,
+                     dtype_t[:] scaling_factors,
+                     dtype_t min_scaling=1e-10):
+    """
+    Compute the alpha probabilities using scaling_factors:
+        P(O_1, O_2, ..., O_t, q_t=S_i | model)
+    """
+    cdef:
+        Py_ssize_t t, i, j
+        dtype_t alpha_t
```
anntzer, line 83: I think what you call alpha here (consistently with the literature) is exp(fwdlattice) in the codebase?  let's stick to fwdlattice for now, we can always do a mass rename later (although I guess it may be a bit annoying that sometimes it includes the log and sometimes it doesn't).

lib/hmmlearn/base.py
blckmaxima: I can do that if you would like.  The reason I went like this is that `frameprob` will be either probabilities, or log-probabilities, depending on which algorithm was used.  I imagined that being explicit with what is happening could be clearer.

If I undo this change, and bring back the single function, do you have a preference for the `frameprob` parameter, knowing that it can have two different interpretations based upon where it is called from ?

anntzer: Ah, I see, that the same problem as fwdlattice vs. alpha as noted below.  I guess we could call it `lattice` to go together with `fwdlattice` and `bwdlattice`?  Probably worth adding a comment somewhere noting that lattice/fwdlattice/bwdlattice correspond either to p/alpha/beta or to logp/logalpha/logbeta depending on implementation.

lib/hmmlearn/base.py
blckmaxima: This should be fixed now.

lib/hmmlearn/_hmmc.pyx
blckmaxima: I've done my best to catch as many of these as possible.

examples/benchmark.py
blckmaxima: This should all be done now.

lib/hmmlearn/base.py
blckmaxima: I've extended scoring to use scaling / log as requested by the user.

lib/hmmlearn/base.py
blckmaxima: OK, I have adjusted this interface - I resisted the urge to make all the names of the parameters in the derived classes be named the same.  We can save that for another day.

I'm open to further suggestions here.  Each of the fit routines could have called `accumulate_sufficient_statistics`, or they could return the needed data to have one call in `fit()`.  I chose the latter option, but could just as easily done the former.

lib/hmmlearn/tests/test_base.py
blckmaxima: Good catch.  I have undid this change and wrapped calls into Cython where needed with `np.asarray`

lib/hmmlearn/_hmmc.pyx
blckmaxima: Fair Enough - Would be happy to help with making both algorithms match the naming in literature in the future.

lib/hmmlearn/base.py
blckmaxima: :+1:

lib/hmmlearn/base.py
blckmaxima: :+1:

blckmaxima:
@anntzer I'll look at the aarch64 builds a bit later.

lib/hmmlearn/_hmmc.pyx
```diff
@@ -64,12 +69,79 @@ def _forward(int n_samples, int n_components,
 
                 fwdlattice[t, j] = _logsumexp(work_buffer) + framelogprob[t, j]
 
+def _forward_scaling(int n_samples, int n_components,
+                     dtype_t[:] startprob,
+                     dtype_t[:, :] transmat,
+                     dtype_t[:, :] frameprob,
+                     dtype_t[:, :] fwdlattice,
+                     dtype_t[:] scaling_factors,
+                     dtype_t min_scaling=1e-90):
```
anntzer, line 78: Is there a specific reason for this value?  If you really want to introduce a check (perhaps just allowing underflow is fine), it would make sense to allow going to the limit of double resolution, i.e. 1e-300.

lib/hmmlearn/tests/test_multinomial_hmm.py
```diff
@@ -13,117 +13,140 @@ class TestMultinomialAgainstWikipedia:
...
-        assert np.allclose(posteriors, [
-            [0.23170303, 0.76829697],
-            [0.62406281, 0.37593719],
-            [0.86397706, 0.13602294],
-        ])
+        for impl in self.implementations:
+            h = self.new_hmm(impl)
+            state_sequence = h.predict(X)
+            posteriors = h.predict_proba(X)
+            print(posteriors)
```
anntzer, line 55: remove prints

lib/hmmlearn/_hmmc.pyx
anntzer: Yes, that would be great as a followup, but let's get this in first.

lib/hmmlearn/base.py
anntzer: I think the current approach is fine for now.

anntzer:
Just a few nits, plus indeed the aarch64 issues (which may be floating point inaccuracy problems, which are always super annoying :-(); otherwise this looks really good :)

lib/hmmlearn/base.py
```diff
@@ -596,6 +678,21 @@ def _compute_log_likelihood(self, X):
...
+        ----------
+        X : array-like, shape (n_samples, n_features)
+            Feature matrix of individual samples.
+
+        Returns
+        -------
+        logprob : array, shape (n_samples, n_components)
+            Log probability of each sample in ``X`` for each of the
+            model states.
+        """
```
anntzer, line 694: Minor nit: this could go above _compute_log_likelihood (likewise in the subclasses), so that the methods related to the scaling version always comes first.

Also, you could do add an implementation like
```python
if self._compute_log_likelihood != _BaseHMM._compute_log_likelihood.__get__(self):  # prevent recursion
    return np.exp(self._compute_log_likelihood(X))
else:
    raise NotImplementedError("Must be overridden in subclass")
```
and likewise to _compute_log_likelihood; this way, subclasses only need to implement *one* of _compute_likelihood and _compute_log_likelihood, whichever is more practical (the overhead from the extra function call should be negligible).

This should also be mentioned at the very end of tutorial.rst, where overridable methods are listed.

blckmaxima: Fair suggestion.  I was planning to prepare a second MR to implement the non-log likelihood methods.  I can do this for now.

lib/hmmlearn/_hmmc.pyx
blckmaxima: I'll update the value, and I have a test I can add to demonstrate how underflow can occur with the scaling algorithm.

lib/hmmlearn/base.py
```diff
@@ -641,8 +747,82 @@ def _initialize_sufficient_statistics(self):
...
+            Probabilities OR Log Probabilities of each sample
+            under each of the model states.  Depends on the choice
+            of implementation of the Forward-Backward algorithm
+
+        posteriors : array, shape (n_samples, n_components)
+            Posterior probabilities of each sample being generated by each
+            of the model states.
+
+        fwdlattice, bwdlattice : array, shape (n_samples, n_components)
+            Log-forward and log-backward probabilities.
```
anntzer, line 773: I think these are "scaled probabilities" for the scaled implementation?

lib/hmmlearn/base.py
```diff
@@ -641,8 +747,82 @@ def _initialize_sufficient_statistics(self):
...
+            "scaling": self._accumulate_sufficient_statistics_scaling,
+            "log": self._accumulate_sufficient_statistics_log,
+        }[self.implementation]
+
+        return impl(stats=stats, X=X, lattice=lattice, posteriors=posteriors,
+                    fwdlattice=fwdlattice, bwdlattice=bwdlattice)
+
+    def _accumulate_sufficient_statistics_scaling(self, stats, X, lattice,
+                                                  posteriors, fwdlattice, bwdlattice):
+        """Updates sufficient statistics from a given sample.
```
anntzer, line 786: I would just say "Implementation of \`_accumulate_sufficient_statistics\` for \`\`implementation = "scaling"\`\`." and not bother with reproducing the docstring (which can otherwise easily go out of sync).  ditto for log.

lib/hmmlearn/_hmmc.pyx
```diff
@@ -86,6 +158,28 @@ def _backward(int n_samples, int n_components,
...
+    """
+    cdef:
+        Py_ssize_t t, i, j
+
+    bwdlattice[:] = 0
+    bwdlattice[n_samples - 1, :] = scaling_factors[n_samples - 1]
+    for t in range(n_samples -2, -1, -1):
+        for j in range(n_components):
+            for i in range(n_components):
+                bwdlattice[t, j] = bwdlattice[t, j] + transmat[j, i] * frameprob[t+1, i] * bwdlattice[t + 1, i]
```
anntzer, line 180: Please linewrap to ~80chars (throughout).

blckmaxima:
Still working on the aarch64 failures - in my local docker / aarch64 setup based on what is used here, all tests are passing for me.  From looking at the specific test, the log-likelihood converges quickly, and then stays the same.  The called function even has a comment about the brittleness.  I'll keep poking at it, something must be different about my local setup.

anntzer:
Perhaps you can do something like
```patch
diff --git i/lib/hmmlearn/tests/__init__.py w/lib/hmmlearn/tests/__init__.py
index 896b09b..ef8899b 100644
--- i/lib/hmmlearn/tests/__init__.py
+++ w/lib/hmmlearn/tests/__init__.py
@@ -34,7 +34,7 @@ def normalized(X, axis=None):
     return X_copy
 
 
-def log_likelihood_increasing(h, X, lengths, n_iter):
+def assert_log_likelihood_increasing(h, X, lengths, n_iter):
     h.n_iter = 1        # make sure we do a single iteration at a time
     h.init_params = ''  # and don't re-init params
     log_likelihoods = np.empty(n_iter, dtype=float)
@@ -45,4 +45,4 @@ def log_likelihood_increasing(h, X, lengths, n_iter):
     # XXX the rounding is necessary because LL can oscillate in the
     #     fractional part, failing the tests.
     diff = np.round(np.diff(log_likelihoods), 10) >= 0
-    return diff.all()
+    assert diff.all(), f"Non-increasing log-likelihoods: {log_likelihoods}"
```
and likewise adjust all the asserts in the codebase; this way it is easier to see how whether the problem is just floating point slop.  (Also, if you do that as a separate PR, I can merge it first, this way you won't have to wait for me to approve each and every one of your pushes on *this* PR for CI to run :-))

anntzer:
You'll need to rebase.

anntzer:
OK, I see the difference is indeed tiny, and it arises from fitting a sequence directly sampled from the HMM (so it makes sense that convergence is really fast and exact).  I think it would make sense to change assert_log_likelihood_increasing to `assert diff.max() < np.finfo(float.eps) ** (1/2)` (i.e. a threshold of around 1.5e-8).  Certainly rounding before checking doesn't make much sense (well, that basically means you allow a threshold of 5e-11).  I don't have any specific reason to believe that sqrt(eps) is the correct threshold, but that seems to be as good a guess as any (and that value occurs reasonably often in similar circumstances, so...).

anntzer:
This looks great. Do you want to squash the commits yourself or should I do it?

blckmaxima:
Squashed!

Issue #340, opened 2019-05-09
Tuple of observations as 1 input observation      


egertteesaar:
Is there a way to give a tuple or a list of observations as 1 observation into MultinomialHMM?

I have this data where there are usually multiple discrete observations for each day but I'd like to describe them with just one hidden state. Can it be done?

anntzer:
No, but I think this is basically the problem at https://github.com/hmmlearn/hmmlearn/issues/335: MultinomialHMM is not really a multinomial HMM but really a categorical one.

Issue #335, opened 2019-04-10
A suggestion about the name "MultinomialHMM"


zhaoyanthu:
This is my personal opinion. I think the name "MultinomialHMM" is not an appropriate name. Why not just call it "DiscreteHMM"? When I first saw "MultinomialHMM", I thought the emission probability supports a general [Multinomial distribution](https://en.wikipedia.org/wiki/Multinomial_distribution).

blake5634:
+1

anntzer:
The distribution implemented by MultinomialHMM is actually a multinomial, isn't it?

wkopp:
It is a categorical distribution that is implemented in MultinomialHMM, which is a special case of a multinomial [Categorical distribution](https://en.wikipedia.org/wiki/Categorical_distribution). I would agree to renaming it. Maybe to CategoricalHMM?

anntzer:
Oh, I see, this is correct.  I'm fine with renaming to CategoricalHMM.

anntzer:
Closed by #432.

PR #429, opened 2021-11-11
Rename MultinominalHMM to CategoricalHMM

trangham283:
Renamed Multinomial HMM functions and test file to Categorical HMM, addressed #335 and #340


trangham283:
Sorry, I messed up the branching, working on a fix now.

blckmaxima:
Does anything in the `docs/source/` directory need to be updated as well ?

trangham283:
Good point, I've updated the docs, and made a note in CHANGES.rst as well.

anntzer:
I saw that you have some other PRs in the works for actually implementing a MultinomialHMM; it may help if you can push them (e.g. as a separate PR) so that we can decide on the best deprecation path for the "old" MultinomialHMM.  In particular, should it switch to one-hot encoding (consistently with the other HMMs, and removing the need for the shape comment in DOC_SUFFIX)?  Even if we agree that a change in behavior is best here, how to make sure that no one accidentally gets "wrong" results due to the semantic changes?  Both questions would be easier to answer if we can have a look at what the new functionality (of the "new" MultinomialHMM) is going to be.

trangham283:
Yes, I implemented "new" MultinomialHMM, but still need to work on some self-contained test examples. I'll submit a separate PR when that is done, unless you think you want to review the implementation sooner than later.

anntzer:
To clarify the problem at hand: The problem is that there are users who currently have code that uses the old MultinomialHMM, and will run e.g. `pip install --upgrade hmmlearn` (upgrading their version of hmmlearn) without necessarily checking the changelog.  It is not acceptable that this *silently* gives them a different result just because you changed what MultinomialHMM means.  Instead, they should either get a DeprecationWarning period, or a noisy crash that tells them, "OK, something changed underneath, I need to keep using the old version of hmmlearn until I can adapt to the new version".

Additionally, it would be better if they don't have to change *twice* their code (once when the old MultinomialHMM goes away, and once when the new one comes in), but just once.  But the best design to achieve this will depend on the new MultinomialHMM API, hence my preference in having at least some draft PR where we can discuss API choices.

trangham283:
The way it's written now, calling MultinomialHMM with features set up for CategoricalHMM will crash, because CategoricalHMM expects (n_samples,) whereas MultinomialHMM expects (n_samples, n_features). 

Ideally, one can use CategoricalHMM with by setting n_trials = 1 in MultinomialHMM, since it is the special case. In that case, the user will get a warning about number of trials as a required parameter (in addition to checking for the shape of the input data). 

In any case, "old" MultinomialHMM is actually not implementing a correct multinomial distribution, which I think is a severe error because it's potentially propagating the wrong information to naive users about what the underlying model is.

trangham283:
Upon further thought, maybe a less ideal workaround but avoids confusion completely is to use "MultinomHMM" (or some other descriptive name) as the class name, getting rid of "MultinomialHMM" completely.

anntzer:
> I think is a severe error

I certainly understand your point, and this may have confused a fair number of people who may or may not have written code using the current multinomial (i.e., categorical).  However, we can reasonably assume that anyone who actually did write this code understood that this is indeed a categorical distribution (regardless of the API name), and we should not gratuitiously break their code.

> CategoricalHMM expects (n_samples,) whereas MultinomialHMM expects (n_samples, n_features).

Thanks, this is the reason why I asked to see the new MultinomialHMM API.

Actually CategoricalHMM also accepts (n_samples, 1), which *could* be confused with a MultinomialHMM with n_features=1; but fortunately that is mostly nonsensical.  Therefore, that means that I am OK with the renaming, as long as MultinomialHMM rejects the n_features=1 case (at least initially; we can always restore support for it later).

> use "MultinomHMM"

Let's not do that, that's not going to help the API clarity.

---

I think in that case, I am *mostly* happy with this PR.  You need to add a clearer entry in the changelog ("MultinomialHMM has been renamed to CategoricalHMM.  There is a new MultinomialHMM class, which implements a multinomial distribution (i.e., repeated samplings from a fixed categorical distribution)."), and I would like to see the *other* (new MultinomialHMM) PR at least in a state where I'd be convinced it would be mergeable at some point in the future (so that the disruption for end users is actually worth it).

trangham283:
Thanks for the detailed feedback, that sounds good to me.

trangham283:
Closed this since all the issues are addressed in #432

PR #431, opened 2021-11-12
Use numpy array operations to accumulate Multinomial emissions statistics`

blckmaxima:
This provides a a significant performance improvement - ~3x for the scaling implementation, 2x for the Logarithm implementation.

### DIFF
--- a/lib/hmmlearn/hmm.py
+++ b/lib/hmmlearn/hmm.py
@@ -479,8 +479,7 @@ class MultinomialHMM(_BaseHMM):
         super()._accumulate_sufficient_statistics(
             stats, X, lattice, posteriors, fwdlattice, bwdlattice)
         if 'e' in self.params:
-            for t, symbol in enumerate(np.concatenate(X)):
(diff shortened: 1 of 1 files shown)


anntzer:
Good ideas, although I have a bit of trouble reproducing the speedups using the benchmark script.
However, I guess the following could/should be even faster (assuming I got my indices correct)?
```python
        if 'e' in self.params:
            np.add.at(stats['obs'].T, np.concatenate(X), posteriors)
```

blckmaxima:
Thank you, I learned something new.  In my testing they are more-or-less equivalent.  I think the numpy version you shared is a bit more idiomatic.  Would you like to make that change?

I will share a separate MR that makes the benchmarks reproducible by optionally making the `random_state` consistent across runs, and ensuring that the specified number of iterations are reached.  This makes measurement a bit more stable.

anntzer:
Please directly integrate my suggested change in this MR (and squash).

Issue #58, opened 2015-11-06
Please clarify the format of priors


anntzer:
`GaussianHMM` provides the `startprob_prior` and `transmat_prior` kwargs, but it isn't clear to me how I'd use them to specify that (in a two state model) I have a prior e.g. of `a_11,a_12~Dirichlet(0.5)` or something else.

Also, is it possible to specify priors for the means and variances?

Thanks.

superbobry:
Hello. Currently `hmmlearn` implement HMMs in a frequentist way, so the (slightly misleading) name *prior* simply means regularizer. I'm not sure if it corresponds to Dirichlet distribution mean.

Similar "priors" exist for means and variances, but I think I forgot to mention them in the docs, so please take a look at the [constructor signature](https://github.com/hmmlearn/hmmlearn/blob/master/hmmlearn/hmm.py#L116).

anntzer:
Thanks you.  I would guess that this indeed corresponds to the Dirichlet prior for starting and transition probabilities.  I need to stare longer at the formulas for regularizing the means and variances to figure out what prior they correspond to -- can you help me there?

jan-matthis:
The prior on the transition matrix/starting probability is indeed a Dirichlet (you can think of it as specifying pseudocounts).

To have a prior on the mean and covariance of a multivariate Gaussian, you would use a normal(-inverse) Wishart distribution. The M-step in `GaussianHMM` is based on this, however, you do not specify the parameters of the prior distribution but rather the variables `covars_prior`, `covars_weight`, `means_prior`, and `means_weight`.  If you want to know how these are related to the parameters of the normal(-inverse) Wishart, you can go about it by deriving the MAP solution and  relating it to the parametrization in the package. 

You might find it helpful to look at: _Maximum A Posteriori Estimation for Multivariate Gaussian Mixture Observations of Markov Chains_ by Gauvain and Lee (1993).

anntzer:
Thanks for the reference.

jexner:
@jan-matthis  This sounds like `startprob_prior` etc. only concern the MAP algorithm, but from the code it looks like it also affects the viterbi algorithm. Are there any other reference on this topic? What would be a typical value range for `startprob_prior`? (I guess it's  not a probability, since then `startprob_` might become negative, see https://github.com/hmmlearn/hmmlearn/blob/master/hmmlearn/base.py#L643.)

jexner:
[...]

PR #432, opened 2021-11-15
Multinomial Implementation

trangham283:
MultinomialHMM class is written in hmm.py, but tests are not finished/just have stubs for now.


blckmaxima:
Hello @trangham283!  Just checking to see if you want to keep working on this ?

trangham283:
Hi @blckmaxima, sorry for dropping the ball on this. Yes, I'd still like to contribute. I've been busy with other projects, and having issues setting up unit tests properly for this, so I apologize for the delay. I'm also a bit confused with all the new recent changes since I haven't caught up with the latest updates, so if it's easier to close this PR and for me to submit a fresh one let me know.

blckmaxima:
You should be able to rebase this on the latest changes...  Let me know if you want help?

trangham283:
Hi @blckmaxima 
Thanks for offering to help! I guess I do need help after all.

I've fixed the conflicts on my end and the implementation for MultinomialHMM is there now. However, I'm having trouble testing it, mainly because [Line 11 in base.py](https://github.com/hmmlearn/hmmlearn/blob/a401dfa854a111138f95e368cd0441dc266ca2c0/lib/hmmlearn/base.py#L11) tries to import _hmmc, which has been moved to src. 

Is there an implicit import for it somewhere? Right now I'm just trying to run pytest and that's the error that comes up. 

Thanks a lot!

anntzer:
If using an editable install (or PYTHONPATH manipulations), have you tried rebuilding the library, e.g. with `python setup.py build_ext -i`?

trangham283:
Thank you! I did try rebuilding, but turns out it was a different problem (my merging problems).

trangham283:
Hi @anntzer @blckmaxima (and maybe @alexrockhill )

Thanks again for all the help & feedback! I think this is ready for you to review once you have time. The tests passed, and I also added warnings regarding the usage of MultinomialHMM. Tagging @alexrockhill since I saw you have an example using (old) MultinomialHMM . 

I added an example as well, to illustrate how it's used (and actually how I used it in my projects).

anntzer:
Can you rebase to resolve the conflicts, or do you need help to do so?

trangham283:
The conflicts are mainly from renaming Multinomial to Categorical, so I don't think I should rebase in this case?

blckmaxima:
@trangham283 We still need to resolve the conflicts before merging :-) - And yes, it looks like they are mainly around renaming to categorical.

trangham283:
Thanks for taking a look @blckmaxima So is there anything I should do on my end? Like I commented earlier, I don't think I should rebase in this case.

trangham283:
I got some help to clean up and rebase. Sorry about all the confusion!

lib/hmmlearn/hmm.py
```diff
@@ -307,18 +309,246 @@ def _do_mstep(self, stats):
                                      (cvweight + stats['post'][:, None, None]))
 
 
-_MULTINOMIALHMM_DOC_SUFFIX = """
+
+class MultinomialHMM(BaseHMM):
+    """
+    Hidden Markov Model with multinomial emissions, e.g. BoW per state
```
anntzer, line 315: what is "BoW"?

lib/hmmlearn/hmm.py
```diff
@@ -307,18 +309,246 @@ def _do_mstep(self, stats):
...
+            normalize(self.transmat_, axis=1)
+        
+        if isinstance(self.startprob_prior, np.ndarray):
+            self.startprob_ = self.startprob_prior.copy()
+        else:
+            self.startprob_ = np.random.random(self.n_components)
+            normalize(self.startprob_)
+
+        if isinstance(self.emission_prior, np.ndarray):
+            self.emissionprob_ = self.emission_prior.copy()
```
anntzer, line 477: copying emission_prior to emissionprob seems wrong, as this is just duplicating the functionality of directly setting the emissionprob attribute, which remains the only approach that works for other hmms (this is also in fact a source of confusion for some users).  I guess I would drop emission_prior for now, or perhaps, if you want a way to specify priors, these should be understood as "pseudocounts" added to the normal counts? (as in https://github.com/hmmlearn/hmmlearn/issues/58#issuecomment-155572844)

lib/hmmlearn/hmm.py
```diff
@@ -307,18 +309,246 @@ def _do_mstep(self, stats):
...
+        if n_features != self.n_features:
+            raise ValueError("Feature dimension should be (n_samples, n_features)")
+        if not (X.sum(axis=1) == self.n_trials).all():
+            raise ValueError("Total count for each sample should add up to number of trials")
+        if not np.issubdtype(X.dtype, np.integer):
+            raise ValueError("Symbols should be integers")
+        if X.min() < 0:
+            raise ValueError("Symbols should be nonnegative")
+        if hasattr(self, "n_features"):
+            if self.n_trials < X.max():
```
anntzer, line 451: I think this condition redundant as it is already checked by `X.sum(-1) == n_trials and all(X >= 0)` ?

lib/hmmlearn/hmm.py
```diff
@@ -307,18 +309,246 @@ def _do_mstep(self, stats):
...
+        random_state = check_random_state(random_state)
+        sample = multinomial.rvs(n=self.n_trials, 
+                p=self.emissionprob_[state,:],
+                size=1, random_state=self.random_state)
+        # sample is of shape (1, n_features); squeeze to keep dimensions consistent
+        return sample.squeeze(0)
+
+    def _initialize_sufficient_statistics(self):
+        stats = super()._initialize_sufficient_statistics()
+        stats['post'] = np.zeros(self.n_components)
```
anntzer, line 521: I guess you don't actually need to accumulate `post`, as `obs` has all the relevant info to normalize itself anyways?

lib/hmmlearn/hmm.py
trangham283: BoW is "bag of words" but I removed it since other HMM description don't go into these specifics.

lib/hmmlearn/hmm.py
trangham283: Makes sense, edited.

lib/hmmlearn/hmm.py
trangham283: Thanks for the reference, I did indeed misunderstand the purpose of emission_prior.

lib/hmmlearn/hmm.py
trangham283: You're right; I took this out and re-tested.

lib/hmmlearn/hmm.py
```diff
@@ -307,18 +309,215 @@ def _do_mstep(self, stats):
...
+                 params="ste", init_params="ste",
+                 implementation="log"):
+        """
+        Parameters
+        ----------
+        n_components : int
+            Number of states, required
+            
+        n_trials : int
+            Number of trials, required; for now assume all samples have the same n_trials
```
anntzer, line 352: Do you plan the change the assumption that n_trials is always the same?  Is there even a "standard" definition for multinomial hmms where n_trials can change?  (e.g. is n_trials also part of the state?)

lib/hmmlearn/hmm.py
```diff
@@ -307,18 +309,215 @@ def _do_mstep(self, stats):
...
+
+    emissionprob_ : array, shape (n_components, n_features)
+        Probability of emitting a given symbol when in each state.
+
+    Examples
+    --------
+    >>> from hmmlearn.hmm import MultinomialHMM
+    """
+
+    def __init__(self, n_components, n_trials, n_features=2,
```
anntzer, line 339: Should n_features be an init arg?  (Does defaulting to 2 really help?)  Should it not be inferred from e.g. the shape of emissionprob (if set) or the fitted time series, basically entirely form _check_and_set_n_features?

lib/hmmlearn/hmm.py
```diff
@@ -307,18 +309,215 @@ def _do_mstep(self, stats):
...
+        Check if ``X`` is a sample from a multinomial distribution, i.e. an
+        array of non-negative integers, summing up to n_trials.
+        """
+        n_samples, n_features = X.shape
+        if n_features != self.n_features:
+            raise ValueError("Feature dimension should be (n_samples, n_features)")
+        if not (X.sum(axis=1) == self.n_trials).all():
+            raise ValueError("Total count for each sample should add up to number of trials")
+        if not np.issubdtype(X.dtype, np.integer):
+            raise ValueError("Symbols should be integers")
```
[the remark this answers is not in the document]
trangham283: yes, entries are symbol counts, fixing this now.

lib/hmmlearn/hmm.py
trangham283: I see what you're saying now. For some reason I thought the other HMMs also had n_features as an init param (maybe because of the comment in this [line](https://github.com/hmmlearn/hmmlearn/blob/1c732510f2e2afa40bc6e1ed3a387cbccff8bbdd/lib/hmmlearn/hmm.py#L43)), but I'm looking at them and that's not the case. So yeah, I'll remove this param.

lib/hmmlearn/hmm.py
trangham283: Originally I intended to allow this, but I'll have to look more carefully to see how big of a change to the code structure it would be, or if something like that would even be sensible. 

As to whether there's a "standard" definition for multinomial hmms, I don't know. But from my understanding, "multinomial HMM" = "an HMM whose emission probabilities are parameterized by multinomial distributions", so by that definition something like that should be allowed. However, implementation/practice-wise, it's not clear to me if that makes sense, especially if each sample to learn from has a different n_trials (and it would also be harder to check valid input samples).

I think it's also possible to have n_trials be different for each component, but keep them the same for all emissions of that component.

blckmaxima:
@trangham283 I ran the pipeline -a few style checks (line length) to check.

trangham283:
Thanks @blckmaxima ! I'm fixing formatting issues now; please ignore the most recent force-push (was trying to clean up again).

lib/hmmlearn/hmm.py
```diff
@@ -307,18 +309,218 @@ def _do_mstep(self, stats):
...
+
+    emissionprob_ : array, shape (n_components, n_features)
+        Probability of emitting a given symbol when in each state.
+
+    Examples
+    --------
+    >>> from hmmlearn.hmm import MultinomialHMM
+    """
+
+    def __init__(self, n_components, n_trials,
```
anntzer, line 339: Looking at the other classes, I think it would be consistent to default n_components=1; this also requires n_trials to have a default value which I would suggest to be None, meaning that it will get auto-set in _check_and_set_multinomial_n_features_n_trials, i.e. an expanded version of _check_and_set_multinomial_n_features which also correctly sets n_trials if it is currently None, or errors out if it is set to an incompatible value.

trangham283: Is there any reason you'd prefer the setting of n_trials to be in a separate function _check_and_set_multinomial_n_features_n_trials instead of doing this in _check_and_set_multinomial_n_features ?

anntzer: It's just a matter of renaming the function so that its name matches what it does?  Or am I missing something?

trangham283: Oh, I thought you meant defining another function `_check_and_set_multinomial_n_features_n_trials` in addition to `_check_and_set_multinomial_n_features`

anntzer:
There are still minor stylistic issues left (as well as a changelog entry missing) but perhaps we should just merge this and fix things after so as not to block the variational inference (and other) PRs.  Thoughts @blckmaxima @alexrockhill?

blckmaxima:
I'm good to merge.  Regarding stylistic issues - I'd propose adopting Black, or some other strict style guide - I definitely need it for my own work!  I also have been moving repos at work to poetry, so perhaps it could be part of that in the future.

anntzer:
I very strongly dislike black (see https://github.com/matplotlib/matplotlib/issues/18796 for some examples of why), but I also realize I'm not actively maintaining this project anymore so if you want to go through the bother of doing so, please go ahead.

(FWIW my personal preference would probably something like adding a flake8 lint and require manual fixing.)

Also pinging @alexrockhill on this PR and on this point.

blckmaxima:
> I very strongly dislike black (see [matplotlib/matplotlib#18796](https://github.com/matplotlib/matplotlib/issues/18796) for some examples of why), but I also realize I'm not actively maintaining this project anymore so if you want to go through the bother of doing so, please go ahead.
> 
> (FWIW my personal preference would probably something like adding a flake8 lint and require manual fixing.)
> 
> Also pinging @alexrockhill on this PR and on this point.

That's great a great thread, thank you for the context.    I can get tons of mileage by setting up a reasonable flake8rc.

anntzer:
I allowed myself to rebase this and clean up some remaining issues to merge this.

PR #433, opened 2021-11-15
Switch cython APIs to return new arrays.

anntzer:
I think this is a nicer API to work with (afaict there isn't any advantage in preallocating the arrays from outside cython rather than inside).  @blckmaxima any thoughts on this?

### DIFF
--- a/lib/hmmlearn/_hmmc.pyx
+++ b/lib/hmmlearn/_hmmc.pyx
@@ -45,10 +45,10 @@ cdef inline dtype_t _logaddexp(dtype_t a, dtype_t b) nogil:
 
 
 def _forward_log(int n_samples, int n_components,
-             dtype_t[:] log_startprob,
-             dtype_t[:, :] log_transmat,
-             dtype_t[:, :] framelogprob,
-             dtype_t[:, :] fwdlattice):
+                 dtype_t[:] log_startprob,
+                 dtype_t[:, :] log_transmat,
+                 dtype_t[:, :] framelogprob,
+                 dtype_t[:, :] fwdlattice):
     """
     Compute the fwdlattice (alpha in the literature)
     probabilities using logarithms:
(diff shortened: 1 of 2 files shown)


blckmaxima:
1. It is easier to read, definitely!
2. In a separate implementation I have (that I am slowly incorporating pieces into here),  I actually allocate these arrays once during `fit()` to be the maximum size needed, and pass them into the forward-backward algorithm.  So, in the multisequence case, if I had sequences of length [100, 500, 1000, 300], I would perform one allocation of size 1000 for the forward/backward lattices, and each pass would reuse as much of the buffer that it needed.  One downside is that you do have to pass these around a bit more :-)  The code I have was used for training ~1000s of sequences, so the memory allocations were noticeable.  


What are the goals for the project?  If the goal is readability, then (1) makes sense.  However, perhaps (2) can be done in a non-messy way?  What do you think?

anntzer:
I certainly value readability, though preferably not at the cost of significant slowdowns.
In the case you mention, have you actually timed that the repeated allocations show up in the timing?  (If so, can you share a branch where the timings are reproducible?)  My understanding is that repeatedly malloc()ing and free()ing blocks of similar sizes should be quite fast (because later mallocs will typically just reuse the just-freed blocks), and the usage pattern you mention certainly seems to match that case (*possibly* it may help to reorder the sequences in decreasing lengths, which is something that can be done with a one-time cost), but I may certainly be wrong...

blckmaxima:
I don't think it will be significant enough to matter.  Unfortunately, I don't have a branch I can point you to - I didn't keep great track of things like this, and the setup was a little different.  The entirety of the equivalent of `fit()` was implemented in Cython.

:+1: from me on this.

anntzer:
OK, let's get this in, I guess we can always discuss different designs if the allocations turn out to be a bottleneck.

PR #434, opened 2021-11-15
Make examples/benchmark.py have reproducible runs

blckmaxima:
If a random-state is specified, pass it through to the models.  Useful for reproducible benchmarks.  Also, require that the minimum number of iterations are reached during training.


anntzer:
It would probably be a good idea to also switch to using https://docs.python.org/3/library/time.html#time.perf_counter (see e.g. https://docs.python.org/3/library/timeit.html#timeit.default_timer).

anntzer:
Sorry this fell through the cracks.  Looks mostly good, needs a rebase though.

Issue #204, opened 2017-05-23
"n_samples=X should be >= n_clusters=X" Error


Dim-P:
As the title says, while I am using Gaussian Mixture Model emissions, I sometimes get errors such as: `n_samples=31 should be >= n_clusters=32`

Here is the full error report:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-94-8a7d39921c58> in <module>()
----> 1 model = GMMHMM(n_components=2,n_mix=32, covariance_type="diag", n_iter=100, algorithm="map").fit(x)
      2 print('GMMHMM converged: ',model.monitor_.converged)
      3 #hidden_states = model.predict(x)
      4 #print(hidden_states)
      5 hidden_states = model.decode(x,algorithm="viterbi")

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/hmmlearn-0.2.1-py3.5-macosx-10.6-intel.egg/hmmlearn/base.py in fit(self, X, lengths)
    422         """
    423         X = check_array(X)
--> 424         self._init(X, lengths=lengths)
    425         self._check()
    426 

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/hmmlearn-0.2.1-py3.5-macosx-10.6-intel.egg/hmmlearn/hmm.py in _init(self, X, lengths)
    616             kmeans = cluster.KMeans(n_clusters=self.n_mix,
    617                                     random_state=self.random_state)
--> 618             kmeans.fit(X[np.where(labels == label)])
    619             kmeanses.append(kmeans)
    620 

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/sklearn/cluster/k_means_.py in fit(self, X, y)
    878         """
    879         random_state = check_random_state(self.random_state)
--> 880         X = self._check_fit_data(X)
    881 
    882         self.cluster_centers_, self.labels_, self.inertia_, self.n_iter_ = \

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/sklearn/cluster/k_means_.py in _check_fit_data(self, X)
    855         if X.shape[0] < self.n_clusters:
    856             raise ValueError("n_samples=%d should be >= n_clusters=%d" % (
--> 857                 X.shape[0], self.n_clusters))
    858         return X
    859 

ValueError: n_samples=31 should be >= n_clusters=32
```

[...]

PR #438, opened 2021-12-01
fixed GMMHMM._init KMeans not constrained by the min cluster size

dizcza:
The issue https://github.com/hmmlearn/hmmlearn/issues/204 was wrongly interpreted and its closure was premature.

Since you do not constrain the k-means clusters to have at least `n_mix` samples, it can easily fit with a smaller no. of samples per cluster. In such cases, when fitting individual k-means, scikit-learn throws an error saying that `n_samples` is less than the `n_clusters` (`n_mix`).

I've fixed that by sampling the means from a normal distribution with zero means and unit variance.

### DIFF
--- a/lib/hmmlearn/hmm.py
+++ b/lib/hmmlearn/hmm.py
@@ -667,13 +667,18 @@ class GMMHMM(_BaseHMM):
         nf = self.n_features
         nm = self.n_mix
 
+        def compute_cv():
+            return np.cov(X.T) + self.min_covar * np.eye(nf)
+
         # Default values for covariance prior parameters
         self._init_covar_priors()
         self._fix_priors_shape()
 
         main_kmeans = cluster.KMeans(n_clusters=nc,
                                      random_state=self.random_state)
+        cv = None  # covariance matrix
         labels = main_kmeans.fit_predict(X)
+        main_centroid = np.mean(main_kmeans.cluster_centers_, axis=0)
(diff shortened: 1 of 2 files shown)


anntzer:
The interpretation seems correct, but I disagree with the fix.  There is no reason why a mean close to zero would make sense (it can be right or also completely wrong, in particular it could could make the likelihood underflow, if the data is really close to zero -- similar underflow problems already occur elsewhere).  It would be more reasonable, I guess(?), to sample the mean from a normal with mean equal to the cluster center and std given by the covariance either input by the user, or computed just below.

dizcza:
How do you see the users inputting the means manually?

> sample the mean from a normal with mean equal to the cluster center and std given by the covariance computed just below.

sounds reasonable even if it comes at a price of computing the covariance matrix in cases when the user has not requested it.

dizcza:
I've added the test that reproduces the old behavior with a fix.

anntzer (approved):

anntzer:
Do you want to squash your commits before merging? (This is up to you.)

dizcza:
Yes, squashing the commits would do just right.

anntzer:
Merged into master by https://github.com/hmmlearn/hmmlearn/commit/4879a8347099c2729896b7742914300e702bbc5d.  Thanks for the PR.

Issue #272, opened 2018-03-19
Feature request: log likelihood of each sequence


remidomingues:
Hello,

The current implementation of HMMs provides quite a few functions returning the likelihood of the data under the trained model such as `predict_proba`, `score` and `score_samples`, but lacks a function which would return the likelihood of each action sequence (instead of a likelihood per sample, or per component). This feature is essential to several applications such as sequence-based anomaly detection.

A first draft of the function could be as follows:

```
def score_samples(self, X, lengths=None)
    check_is_fitted(self, "startprob_")
    self._check()

    X = check_array(X)
    logprob = []
    for i, j in iter_from_X_lengths(X, lengths):
        framelogprob = self._compute_log_likelihood(X[i:j])
        logprob.append(self._do_forward_pass(framelogprob)[0])
    return np.array(logprob)
```

What do you think?

anntzer:
You could just call score_samples independently on each sequence? Unless I am mistaken, there is no specific advantage in scoring everything together?

PR #439, opened 2021-12-06
add score_batches method which returns a list of scores

Dobatymo:
Added `score_batches` method which returns a list of scores (one for each sequence) instead of a cumulative score.

See https://github.com/hmmlearn/hmmlearn/issues/272

### DIFF
--- a/lib/hmmlearn/_hmmc.pyx
+++ b/lib/hmmlearn/_hmmc.pyx
@@ -23,6 +23,8 @@ cdef inline dtype_t _max(dtype_t[:] X) nogil:
 
 cdef inline dtype_t _logsumexp(dtype_t[:] X) nogil:
     cdef dtype_t X_max = _max(X)
+    cdef ssize_t i
+
     if isinf(X_max):
         return -INFINITY
 
--- a/lib/hmmlearn/base.py
+++ b/lib/hmmlearn/base.py
@@ -253,6 +253,64 @@ class _BaseHMM(BaseEstimator):
         """
         return self._score(X, lengths, compute_posteriors=False)[0]
 
+    def score_batches(self, X, lengths=None):
+        """
+        Compute the log probability under the model.
+        Compared to score() it returns a list of probabilities and doesn't add them up.
+
+        Parameters
+        ----------
+        X : array-like, shape (n_samples, n_features)
+            Feature matrix of individual samples.
+        lengths : array-like of integers, shape (n_sequences, ), optional
+            Lengths of the individual sequences in ``X``. The sum of
+            these should be ``n_samples``.
+
+        Returns
+        -------
+        logprobs : List[float]
+            Log likelihoods of sequences in ``X``.
+
+        """
+
+        _utils.check_is_fitted(self, "startprob_")
+        self._check()
+
+        X = check_array(X)
+        impl = {
+            "scaling": self._score_batches_scaling,
+            "log": self._score_batches_log,
+        }[self.implementation]
+        return impl(X=X, lengths=lengths)
+
+    def _score_batches_log(self, X, lengths=None):
+
+        logprobs = []
+        for sub_X in _utils.split_X_lengths(X, lengths):
+            try:
+                framelogprob = self._compute_log_likelihood(sub_X)
+            except IndexError:
+                logprobij = -np.inf
+            else:
+                logprobij, _ = self._do_forward_log_pass(framelogprob)
+            logprobs.append(logprobij)
+
+        return logprobs
+
+    def _score_batches_scaling(self, X, lengths=None):
+
+        logprobs = []
+        for sub_X in _utils.split_X_lengths(X, lengths):
+            try:
+                frameprob = self._compute_likelihood(sub_X)
+            except IndexError:
+                logprobij = -np.inf
+            else:
+                logprobij, _, _ = self._do_forward_scaling_pass(frameprob)
+            logprobs.append(logprobij)
+
+        return logprobs
+
     def _score(self, X, lengths=None, *, compute_posteriors):
         """
         Helper for `score` and `score_samples`.
--- a/lib/hmmlearn/tests/test_base.py
+++ b/lib/hmmlearn/tests/test_base.py
@@ -149,6 +149,22 @@ class TestBaseAgainstWikipedia:
                                   [0.8673, 0.1327]])
         assert np.allclose(posteriors, refposteriors, atol=1e-4)
 
+    def test_score_batches_log(self):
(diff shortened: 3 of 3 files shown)


lib/hmmlearn/_hmmc.pyx
```diff
@@ -23,6 +23,8 @@ cdef inline dtype_t _max(dtype_t[:] X) nogil:
 
 cdef inline dtype_t _logsumexp(dtype_t[:] X) nogil:
     cdef dtype_t X_max = _max(X)
+    cdef ssize_t i
```
anntzer, line 26: unneeded?

lib/hmmlearn/base.py
```diff
@@ -253,6 +253,64 @@ def score(self, X, lengths=None):
         """
         return self._score(X, lengths, compute_posteriors=False)[0]
 
+    def score_batches(self, X, lengths=None):
```
anntzer, line 256: Are there similar APIs in scikit-learn or related packages?  If yes, how are they named?

lib/hmmlearn/base.py
```diff
@@ -253,6 +253,64 @@ def score(self, X, lengths=None):
...
+            Feature matrix of individual samples.
+        lengths : array-like of integers, shape (n_sequences, ), optional
+            Lengths of the individual sequences in ``X``. The sum of
+            these should be ``n_samples``.
+
+        Returns
+        -------
+        logprobs : List[float]
+            Log likelihoods of sequences in ``X``.
+
```
anntzer, line 273: no blank line.

lib/hmmlearn/base.py
```diff
@@ -253,6 +253,64 @@ def score(self, X, lengths=None):
...
+        return impl(X=X, lengths=lengths)
+
+    def _score_batches_log(self, X, lengths=None):
+
+        logprobs = []
+        for sub_X in _utils.split_X_lengths(X, lengths):
+            try:
+                framelogprob = self._compute_log_likelihood(sub_X)
+            except IndexError:
+                logprobij = -np.inf
```
anntzer, line 293: when does this occur?  why is this not covered by tests?

lib/hmmlearn/tests/test_base.py
```diff
@@ -149,6 +149,22 @@ def test_score_samples(self):
                                   [0.8673, 0.1327]])
         assert np.allclose(posteriors, refposteriors, atol=1e-4)
 
+    def test_score_batches_log(self):
+        # ``StubHMM` ignores the values in ```X``, so we just pass in an
+        # array of the appropriate shape.
+        logprobs = self.hmm._score_batches_log(self.framelogprob)
+
+        reflogprob = -3.3725
+        assert round(sum(logprobs), 4) == reflogprob
```
anntzer, line 158: perhaps make a more stringent test?

lib/hmmlearn/base.py
blckmaxima: How about naming this `score_samples`, it is what the GaussianMixtureModel in sklearn provides.

lib/hmmlearn/tests/test_base.py
blckmaxima: Even just a single function that loops over the various `implementations` - that is how I introduced the log/scaling logic.

lib/hmmlearn/base.py
anntzer: The problem is that a "sample" would typically be a single observation (and the confusion typically doesn't occur in sklearn); perhaps score_sequences?  (i.e. implicitly "score the individual sequences")

blckmaxima: That, and there already exists a `score_samples` method ;-) 

+1 for `score_sequences`.

lib/hmmlearn/_hmmc.pyx
Dobatymo: I added it for consistency. In all other functions the iteration variables are declared as well.

lib/hmmlearn/base.py
Dobatymo: `score_sequences` sounds good to me too. There isn't really a similar API in sklearn I think.

lib/hmmlearn/base.py
Dobatymo: It occurs when states not found in training are scored in `MultinomialHMM` for example. In that case it makes sense to have this specific sequence return -inf since it encountered a zero chance transition I think.

ie. `fit([[0])` `score([[1]])`

lib/hmmlearn/_hmmc.pyx
anntzer: Oh, I see.  Can you make that a separate commit to keep things self-contained(ish)?  I can do it directly, too.

lib/hmmlearn/base.py
```diff
@@ -253,6 +253,63 @@ def score(self, X, lengths=None):
...
+        ----------
+        X : array-like, shape (n_samples, n_features)
+            Feature matrix of individual samples.
+        lengths : array-like of integers, shape (n_sequences, ), optional
+            Lengths of the individual sequences in ``X``. The sum of
+            these should be ``n_samples``.
+
+        Returns
+        -------
+        logprobs : List[float]
```
anntzer, line 271: I guess the "modern" syntax is `list[float]` (lowercase).

lib/hmmlearn/base.py
```diff
@@ -253,6 +253,63 @@ def score(self, X, lengths=None):
...
+        self._check()
+
+        X = check_array(X)
+        impl = {
+            "scaling": self._score_sequences_scaling,
+            "log": self._score_sequences_log,
+        }[self.implementation]
+        return impl(X=X, lengths=lengths)
+
+    def _score_sequences_log(self, X, lengths=None):
```
anntzer, line 285: I think you could modify _score_log/_score_scaling to not sum the individual scores (and then sum them in _score), and then you could reuse them here instead of duplicating the logic and having the IndexError specialcase just in score_sequences.

Dobatymo: Would you like all score methods to return -inf instead of IndexError?

anntzer: Returning -inf seems reasonable, as that's also what gets returned with `fit([[1]]); score([[0]])`.  It may be worth mentioning this behavior in the docstring of MultinomialHMM, too.

lib/hmmlearn/base.py
anntzer: thanks for the example, see comment above.

PR #440, opened 2021-12-06
Parallelize `fit` and `score_batches` with joblib

Dobatymo:
Since most of the cython code releases the GIL, the joblib threading backend can be used to parallelize some of the operations. Since a Python dictionary is used after the calculations to collect statistics, I used a threading lock to prevent races.

I am not 100% sure if I missed any other places which would need to be locked however.


lib/hmmlearn/base.py
```diff
@@ -2,11 +2,13 @@
 import string
 import sys
 from collections import deque
+import threading
 
 import numpy as np
 from scipy import linalg, special
 from sklearn.base import BaseEstimator
 from sklearn.utils import check_array, check_random_state
+from joblib import Parallel, delayed
```
anntzer, line 11: Can you use stdlib's concurrent.futures instead, which provides AFAICT a very similar API?

lib/hmmlearn/base.py
```diff
@@ -494,20 +561,23 @@ def fit(self, X, lengths=None):
...
-                # update their probability distributions, so keep
-                # a single call to this method for simplicity.
+        
+        def process_sequence(sub_X, stats):
+            lattice, logprob, posteriors, fwdlattice, bwdlattice = \
+                    impl(sub_X)
+            # Derived HMM classes will implement the following method to
+            # update their probability distributions, so keep
+            # a single call to this method for simplicity.
+            with self.stats_lock:
```
anntzer, line 571: The need for locking here vaguely suggests to me that _accumulate_sufficient_statistics is not, in fact, the right API; rather, we should perhaps have a _summarize_sufficient_statistics that *returns* the info for a chunk and then aggregate the chunks' individual infos in _do_mstep.  Thoughts?
(Perhaps also relevant for @dizcza wrt #437?)

Dobatymo: Yes it would be great to get rid of the lock here. Returning the info and aggregating it later would be one way, especially if the aggregation is order independent. However looking at some of the different HMM implementations, there's quite a lot data involved which might lead to increased memory usage.

dizcza: IMHO the aggregation should be order-independent. But as @Dobatymo pointed out, the unreduced statistics gathered from a batch of samples could be quite heavy in memory usage, and transferring it here and there is not a good idea (again, IMHO). The approach I want to bring on top of #437 is the ability to reduce (to do a summation over the batch dimension) directly in the `_accumulate_sufficient_statistics` function. Therefore, a lock is needed. But after the summation is done - no need to wait while each thread/process does its summation - they are independent. And I agree with @anntzer that we'll benefit from having a `_summarize_sufficient_statistics` function, in which each job reduces the batch it processes (does the summation) and returns the summary that can be efficiently added to the global accumulated statistics.
Just like in CUDA programming.

lib/hmmlearn/_hmmc.pyx
```diff
@@ -23,6 +23,8 @@ cdef inline dtype_t _max(dtype_t[:] X) nogil:
 
 cdef inline dtype_t _logsumexp(dtype_t[:] X) nogil:
     cdef dtype_t X_max = _max(X)
+    cdef ssize_t i
```
dizcza, line 26: I ain't a reviewer, just spotted this in the first line. Do you ever use `i` later on?

Dobatymo:
Since `_accumulate_sufficient_statistics` appends to a list in GMMHMM, is the order information unimportent in m_step later? If not my current solution would be incorrect.

I think another way would be to split the behaviour into collect and aggregate statistics functions. `collect` would return the necessary info from the thread as decribed by @anntzer and `aggregate` would do the aggregation live on the main thread without using batches. I don't think the aggregation should be done in `_do_mstep` since then all stats would need to be saved until then. Then the aggregation would only have to keep up with the calculation threads. I haven't looked into https://github.com/hmmlearn/hmmlearn/pull/437 yet.

I can also rework the PR to use `concurrent.futures.ThreadPoolExecutor` instead of joblib.

dizcza:
> Since _accumulate_sufficient_statistics appends to a list in GMMHMM

There won't be any appends in the _accumulate_sufficient_statistics in the future. Everything will be accumulated there (just like the _do_mstep currently does) to avoid memory growth.

> I don't think the aggregation should be done in _do_mstep since then all stats would need to be saved until then.

Yes, in the future, the _do_mstep will only update the parameters based on already aggregated statistics.

> collect would return the necessary info from the thread as decribed by @anntzer and aggregate would do the aggregation live on the main thread without using batches.

If `collect` reduces over the batch dimension, then it's a good idea.

Dobatymo:
I ran some benchmarks and it doesn't scale great (at least on my laptop 4 cores, 8 with hyper threading)

Master branch: 132.57 seconds

max_workers | Time
-- | --
1 | 125.25s
2 | 79.45
3 | 71.93s
4 | 87.98s
5 | 116.87s
6 | 155.95s
7 | 179.64s
8 | 183.98s

lib/hmmlearn/base.py
```diff
@@ -143,7 +157,7 @@ def __init__(self, n_components=1,
                  n_iter=10, tol=1e-2, verbose=False,
                  params=string.ascii_letters,
                  init_params=string.ascii_letters,
-                 implementation="log"):
+                 implementation="log", max_workers=None):
```
anntzer, line 160: Needs documentation; also, the default should stay "no multiprocessing, no use of threads at all".

lib/hmmlearn/base.py
```diff
@@ -745,16 +822,12 @@ def _initialize_sufficient_statistics(self):
                  'trans': np.zeros((self.n_components, self.n_components))}
         return stats
 
-    def _accumulate_sufficient_statistics(self, stats, X, lattice,
+    def _accumulate_sufficient_statistics(self, X, lattice,
```
anntzer, line 825: I think the change in _accumulate_sufficient_statistics should occur first, and in a separate PR, so that that API can be ironed out.  It will also need some changelog entries so that anyone who implemented custom HMM subclasses knows what they need to change.

lib/hmmlearn/base.py
fcostin: I agree with @anntzer that there must be a way for users of the libary to disable threads and multiprocessing, and this should be the the default for backwards company.

There are various scenarios where someone will be using hmmlearn in a library and have some other solution to achieve parallel or concurrent computation that `hmmlearn` is not and should not be aware of. In such cases, if all cores and threads of a machine are already fully utilised, hmmlearn trying to do its own multithreading will reduce performance.

PR #443, opened 2021-12-21
Switch from cython to pybind11 to workaround setuptools issues.

anntzer:
It seems like the last release of setuptools broke interaction with cython (and also with numpy.distutils, which we don't use but scikit-learn uses to build the py3.10 wheel).  I already had some plans to switch the extension module to pybind11, which has, afaict, fewer awkward interactions with setuptools (it also seems like setuptools is going through some heavy refactoring so we may expect more breakage in the future).  There's at least another objective advantages to pybind11 (the extension module does not depend on the numpy version it was built against, which avoids the need for tricks like https://pypi.org/project/oldest-supported-numpy/), and I tend to prefer c++ to cython anyways.

Just pinging @blckmaxima (who has fiddled with the extension modules recently) here before merging this, in case he has an objection against the switch to pybind11.

Note that CI will still fail on windows + py3.10, but that's when building scikit-learn, which occurs *after* we checked that hmmlearn can be correctly built.


blckmaxima:
Looks good - still seems very readable.

PR #448, opened 2022-01-16
GMMHMM: moved for-loops computations from _do_mstep to _accumulate_sufficient_statistics

dizcza:
This PR completes https://github.com/hmmlearn/hmmlearn/pull/437
The for-loops introduced in the above-mentioned PR were an intermediate step that shed the light on how we can get rid of storing `post_comp_mix` in the stats. Now, these for-loops are removed, and the logic of computing `c_n` is moved to `_accumulate_sufficient_statistics`.
Finally, an old FIXME is resolved.

### DIFF
--- a/lib/hmmlearn/hmm.py
+++ b/lib/hmmlearn/hmm.py
@@ -906,8 +906,11 @@ class GMMHMM(_BaseHMM):
         stats['post_mix_sum'] = np.zeros((self.n_components, self.n_mix))
         stats['post_sum'] = np.zeros(self.n_components)
 
-        lambdas, mus = self.means_weight, self.means_prior
-        stats['m_n'] = lambdas[:, :, None] * mus
+        if 'm' in self.params:
+            lambdas, mus = self.means_weight, self.means_prior
+            stats['m_n'] = lambdas[:, :, None] * mus
+        if 'c' in self.params:
+            stats['c_n'] = np.zeros_like(self.covars_)
 
         # These statistics are stored in arrays and updated in-place.
         # We accumulate chunks of data for multiple sequences (aka
@@ -929,7 +932,7 @@ class GMMHMM(_BaseHMM):
         # Statistics shapes:
         # post_comp_mix     (n_samples, n_components, n_mix)
         # samples           (n_samples, n_features)
-        # centred           (n_samples, n_components, n_mix, n_features)
+        # centered          (n_samples, n_components, n_mix, n_features)
 
         post_mix = np.zeros((n_samples, self.n_components, self.n_mix))
         for p in range(self.n_components):
@@ -944,8 +947,8 @@ class GMMHMM(_BaseHMM):
         stats['post_mix_sum'] += post_comp_mix.sum(axis=0)
         stats['post_sum'] += post_comp.sum(axis=0)
 
-        # means stats
-        stats['m_n'] += np.einsum('ijk,il->jkl', post_comp_mix, X)
+        if 'm' in self.params:  # means stats
+            stats['m_n'] += np.einsum('ijk,il->jkl', post_comp_mix, X)
 
         if 'c' in self.params:  # covariance stats
             centered = X[:, None, None, :] - self.means_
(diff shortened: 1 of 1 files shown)


lib/hmmlearn/hmm.py
```diff
@@ -906,34 +906,16 @@ def _initialize_sufficient_statistics(self):
         stats['post_mix_sum'] = np.zeros((self.n_components, self.n_mix))
         stats['post_sum'] = np.zeros(self.n_components)
 
-        # The following statistics are stored in lists so we can
-        # accumulate chunks of data for multiple sequences (aka
+        lambdas, mus = self.means_weight, self.means_prior
+        stats['m_n'] = lambdas[:, :, None] * mus
```
anntzer, line 910: I think this, as well as the update to m_n in accumulate_sufficient_statistics, only needs to be done `if 'm' in self.params`?

lib/hmmlearn/hmm.py
```diff
@@ -944,7 +926,10 @@ def _accumulate_sufficient_statistics(self, stats, X, lattice,
 
         n_samples, _ = X.shape
 
-        stats['samples'].append(X)
+        # Statistics shapes:
+        # post_comp_mix     (n_samples, n_components, n_mix)
+        # samples           (n_samples, n_features)
+        # centred           (n_samples, n_components, n_mix, n_features)
```
anntzer, line 932: typo: centered

lib/hmmlearn/hmm.py
```diff
@@ -955,11 +940,41 @@ def _accumulate_sufficient_statistics(self, stats, X, lattice,
...
+                centered_norm = norm_last(centered)
+                c_n = np.einsum('ijk,ijk->jk', post_comp_mix, centered_norm)
+            elif self.covariance_type == 'tied':
+                centered_dots = outer_f(centered)
+                c_n = np.einsum('ijk,ijklm->jlm', post_comp_mix, centered_dots)
+
+            if 'c_n' in stats:
+                stats['c_n'] += c_n
+            else:
+                stats['c_n'] = c_n
```
anntzer, line 976: I guess you could init c_n to zeros in initialize_sufficient_statistics to avoid that check.

anntzer:
Mostly looking good for me except for minor points, giving a few days for @fcostin and @blckmaxima in case they want to comment as well.

lib/hmmlearn/hmm.py
dizcza: you're right. done

anntzer (approved):

blckmaxima:
Note, according to https://www.cambridge.org/core/books/bayesian-speech-and-language-processing/94E5BBCCAC14F5AC06BC4C42651D9C57 around page ~156, this can be even further simplified in the future.

fcostin:
nice patch. good to see huge `stats['post_comp_mix']` and `stats['samples']` removed in favour of reducing over the chunks in `_accumulate_sufficient_statistics`.

lib/hmmlearn/hmm.py
```diff
@@ -955,11 +943,38 @@ def _accumulate_sufficient_statistics(self, stats, X, lattice,
...
+            stats['m_n'] += np.einsum('ijk,il->jkl', post_comp_mix, X)
+
+        if 'c' in self.params:  # covariance stats
+            centered = X[:, None, None, :] - self.means_
+
+            def outer_f(x):  # Outer product over features.
+                return x[..., :, None] * x[..., None, :]
+
+            def norm_last(x):  # Much faster than (x**2).sum(-1).
+                return np.einsum('...i,...i', x, x)
```
fcostin, line 960: minor nit-pick:  the definitions of these two functions are now duplicated in `_accumulate_sufficient_statistics` and also in `_do_mstep`.  perhaps it'd be better to define them once in global scope of this file (perhaps renamed as `_outer_f` and `_norm_last`,  at the cost of making the calculations slightly less locally readable

anntzer: I think these are short enough that it's actually OK and more readable to duplicate the implementations to have them readily available (in fact norm_last can even just be inlined...)

anntzer:
Squashed and merged into master as eefb005.  Thanks!

dizcza:
@anntzer as a side note, why do you create a separate commit with squashed commits instead of squashing and merging a pull request? In the end, you will still have a single commit but with the pull request merged, not closed.

anntzer:
Habits are hard to change... I'll remember to try the other approach.

PR #456, opened 2022-03-18
[ENH, MRG] Add decoding example

alexrockhill:
I tried a bit to make a decoding example with simulated data. There aren't any time-series datasets in `sklearn` so I simulated one in the example. I think it's a bit too complicated as is, but even now, there might be seasonal and rain states that are interpretable. I'll work on it a bit more but I thought I would push it as a work in progress to see if anyone wanted to comment.


alexrockhill:
I just simplified the example and added it to the existing one, that's probably easier to start with.

blckmaxima:
Can you call the new file: `plot_hmm_sampling_and_decoding.py` and also update `doc/source/tutorial.txt` to reference this new file?

blckmaxima:
And squash commits!

alexrockhill:
Sure, no problem, done.

examples/plot_hmm_sampling_and_decoding.py
```diff
@@ -0,0 +1,115 @@
...
+             bbox=dict(alpha=.7, facecolor='w'))
+plt.legend(loc='best')
+plt.show()
+
+# %%
+# Now, let's ensure we can recover our parameters.
+
+scores = list()
+models = list()
+rkf = RepeatedKFold(n_splits=2, n_repeats=10, random_state=99)
```
anntzer, line 68: I am confused: how can you get the correct transition matrix after k-fold splitting the input timeseries?

examples/plot_hmm_sampling_and_decoding.py
```diff
@@ -0,0 +1,115 @@
...
+    for train_idx, validate_idx in rkf.split(X):
+        # define our hidden Markov model
+        model = hmm.GaussianHMM(n_components=n_components,
+                                covariance_type='full',
+                                random_state=len(models), n_iter=10)
+        try:
+            model.fit(X[train_idx])
+            models.append(model)
+            scores.append(model.score(X[validate_idx]))
+            print(f'Converged: {model.monitor_.converged}\tScore:{scores[-1]}')
```
anntzer, line 79: Add a space after the second colon.

examples/plot_hmm_sampling_and_decoding.py
```diff
@@ -0,0 +1,115 @@
...
+
+# %%
+# Let's plot our states compared to those generated and our transition matrix
+# to get a sense of our model. We can see that the recovered states follow
+# the same path as the generated states, just with the identities of the
+# states transposed. The same is true for the transition matrix.
+
+# plot model states over time
+fig, ax = plt.subplots()
+ax.plot(Z, states)
```
anntzer, line 100: It took me a bit of time to understand this plot; I agree once you understand it it is convincing but it perhaps warrants a slightly more extended explanation?

examples/plot_hmm_sampling_and_decoding.py
alexrockhill: Ah good point, these should be split once. I was just having trouble random seeding the Baum-Welch algorithm so I did a bit of a hack and changed up the training data but I should just seed it.

alexrockhill: Good point! I just split 50/50. I was doing that because I'm having trouble seeding it: I get the same score every time for the same number of components. The version I just pushed has that issue, do you know how to fix it?

alexrockhill: Ah, okay, I looked into the code a bit more and the `random_state` is used to seed KMeans clustering for the prior on the means. I think maybe let's just start with this and I can add a more complex example that has different KMeans solutions at different random states and show how to seed the EM algorithm in a followup. Probably good to keep the first example simple.

anntzer:
Please revert the pep fixes (and put them as a separate PR if you want to discuss them).

alexrockhill:
> Please revert the unrelated pep fixes (and put them as a separate PR if you want to discuss them; let's try to keep the discussion here focused on the example).

Sure, will do, I was just looking into why the seeding failed (because the means are instantiated by KMeans clustering) and saw somethings but I will do those separately.

anntzer (approved):

anntzer:
I'll give @blckmaxima a couple of days in case he wants to chime in; other than that, this looks like an improvement (we can always refine it later).

blckmaxima:
Thanks for the time - I've been busy with a flooded basement :-)

blckmaxima:
Looks Good!

blckmaxima (approved):

PR #458, opened 2022-03-24
[ENH, MRG] add example with training

### LINKED ISSUE #70: simple multinomial example

alexrockhill:
This goes a bit beyond https://github.com/hmmlearn/hmmlearn/pull/456 since the KMeans clustering makes it so that it has to be a fairly complicated dataset to have different initializations (different clusters based on different starting points) so I just implemented a classic example of a casino with a loaded die. I implemented it exactly as http://www.mcb111.org/w06/durbin_book.pdf but the training process fails to reach a global maximum even with 1000 random starting positions. I'm not sure what's wrong, I'll have to look into it.

Edit by @anntzer: closes #70.

### DIFF
--- a/examples/plot_casino_example.py
+++ b/examples/plot_casino_example.py
@@ -1,0 +1,147 @@
+"""
+Dishonest Casino Example
+------------------------
+
+We'll use the ubiquitous dishonest casino example to demonstrate
+how to train a Hidden Markov Model (HMM) on somewhat realistic
+test data (e.g. http://www.mcb111.org/w06/durbin_book.pdf
+Chapter 3).
+
+In this example, we suspect that a casino is trading out a fair
+die (singular or dice) for a loaded die. We want to figure out
+1) when the loaded die was used (i.e. the most likely path) 2) how
+often the loaded die is used (i.e. the transition probabilities) and
+3) the probabilities for each outcome of a roll for the loaded die
+(i.e. the emission probabilities).
+"""
+
+# %%
+# First, import necessary modules and functions.
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+from hmmlearn import hmm
+
+# %%
+# Now, let's act as the casino and exchange a fair die for a loaded one
+# and generate a series of rolls that someone at the casino would
+# observe.
+
+# make our generative model with two components, a fair die and a
+# loaded die
+gen_model = hmm.MultinomialHMM(n_components=2, random_state=99)
+
+# the first state is the fair die so let's start there so no one
+# catches on rigth away
+gen_model.startprob_ = np.array([1.0, 0.0])
+
+# now let's say that we sneak the loaded die in:
+# here, we have a 95% chance to continue using the fair die and a 5%
+# chance to switch to the loaded die
+# when we enter the loaded die state, we have a 90% chance of staying
+# in that state and a 10% chance of leaving
+gen_model.transmat_ = np.array([[0.95, 0.05],
+                                [0.1, 0.9]])
+
+# now let's set the emission means:
+# the first state is a fair die with equal probabilities and the
+# second is loaded by being biased toward rolling a six
+gen_model.emissionprob_ = \
+    np.array([[1 / 6, 1 / 6, 1 / 6, 1 / 6, 1 / 6, 1 / 6],
+              [1 / 10, 1 / 10, 1 / 10, 1 / 10, 1 / 10, 1 / 2]])
+
+# simulate the loaded dice rolls
+rolls, gen_states = gen_model.sample(30000)
+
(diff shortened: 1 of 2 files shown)


alexrockhill:
I had an issue setting `transmat_prior` which was not setting `model.transmat_` but instead was a parameter in the Dirichlet prior, I should have read more carefully. This works now though. I think this would be really helpful because it shows an actually example where optimization isn't just solved by K-means basically and so shows how to initialize multiple models which I think is important.

examples/plot_casino_example.py
```diff
@@ -0,0 +1,149 @@
...
+fair_to_loaded_priors = rng.random(size=n_fits) / 4
+loaded_to_fair_priors = rng.random(size=n_fits) / 4
+for idx in range(n_fits):
+    model = hmm.MultinomialHMM(
+        n_components=2, random_state=idx,
+        init_params='se')  # don't init transition, set it below
+    # we need to initialize with random transition matrix probabilities
+    # because the default is an even likelihood transition
+    fair_to_loaded = fair_to_loaded_priors[idx]
+    loaded_to_fair = loaded_to_fair_priors[idx]
```
anntzer, line 103: I would just draw the random init here (`fair_to_loaded = 0.25*random()`); there's not advantage of drawing them all at the beginning and indexing.

Also hmmlearn currently doesn't require the "new" numpy random api and I think we should keep using that in the example as well (to avoid having a dependency on numpy 1.17).

anntzer:
Mostly looks good, also giving a few days to @blckmaxima on this one.

examples/plot_casino_example.py
alexrockhill: Ok just using `np.random.random`

alexrockhill:
> Mostly looks good, also giving a few days to @blckmaxima on this one.

Sounds good. My goal was just to add a few more practical examples to demonstrate the API use so that users might not have to figure it out so much on their own. I know that it's in the README that `hmmlearn` is in low maintenance so I don't want to bug anyone too much reviewing PRs. I didn't have anything ambitious in mind but I thought it might be nice to add a `PoissonHMM` and a few things like that. Is that in line with the future of `hmmlearn` or would it be better to table that kind of thing? Again, I completely understand the need for low-maintenance.

anntzer:
> I thought it might be nice to add a PoissonHMM and a few things like that. Is that in line with the future of hmmlearn or would it be better to table that kind of thing? Again, I completely understand the need for low-maintenance.

Please do.  I can't guarantee very fast review times but I'll try to have a look, and if you don't submit a PR things will certainly never happen :)

blckmaxima:
@alexrockhill  Would love a PoissonHMM!

blckmaxima (approved):

alexrockhill:
@blckmaxima, do you have time to look at this one as well?

blckmaxima:
@alexrockhill Would you like to add a reference to this example in the documentation?  And then update the commit message to something like "Add example demonstrating a HMM learning a dishonest casino" (or appropriate)

alexrockhill:
Thanks for the suggestion, I added it as an example under finding a global optimal fit using multiple initializations because that is I think it's most important use case that isn't sufficiently covered.

PR #461, opened 2022-04-04
Variational inference

blckmaxima:
@anntzer  Leaving an early draft of incorporating Variational Inference training of HMMs so I may receive feedback before I keep going.

Some Notes:
* I derive from BaseHMM, and am able to reuse most of it, with a few exceptions.
* `VariationalGaussianHMM` is still incomplete - only Full Covariance is supported.
* Tests are lacking.

Up Next:
* Finish the different covariance types for Gaussian
* Add Mixture of Gaussian Emissions

### DIFF
--- a/lib/hmmlearn/base.py
+++ b/lib/hmmlearn/base.py
@@ -576,6 +576,16 @@ class BaseHMM(BaseEstimator):
             return True
         return False
 
+    def _check_and_set_n_features(self, X):
+        _, n_features = X.shape
+        if hasattr(self, "n_features"):
+            if self.n_features != n_features:
+                raise ValueError(
+                    f"Unexpected number of dimensions, got {n_features} but "
+                    f"expected {self.n_features}")
+        else:
+            self.n_features = n_features
+
     def _get_n_fit_scalars_per_param(self):
         """
         Return a mapping of fittable parameter names (as in ``self.params``)
@@ -595,6 +605,7 @@ class BaseHMM(BaseEstimator):
         X : array-like, shape (n_samples, n_features)
             Feature matrix of individual samples.
         """
+        self._check_and_set_n_features(X)
         init = 1. / self.n_components
         random_state = check_random_state(self.random_state)
         if self._needs_init("s", "startprob_"):
--- a/lib/hmmlearn/base.py
+++ b/lib/hmmlearn/base.py
@@ -4,11 +4,11 @@ import sys
 from collections import deque
 
 import numpy as np
-from scipy import linalg
+from scipy import linalg, special
 from sklearn.base import BaseEstimator
 from sklearn.utils import check_array, check_random_state
 
-from . import _hmmc, _utils
+from . import _hmmc, _kl_divergence as _kl, _utils
 from .utils import normalize, log_normalize
 
 
@@ -52,9 +52,9 @@ class ConvergenceMonitor:
     ...                                   model.monitor_.verbose)
     """
 
-    _template = "{iter:>10d} {log_prob:>16.4f} {delta:>+16.4f}"
+    _template = "{iter:>10d} {log_prob:>16.8f} {delta:>+16.8f}"
 
-    def __init__(self, tol, n_iter, verbose):
+    def __init__(self, tol, n_iter, verbose, *, strict=False):
         """
         Parameters
         ----------
@@ -66,17 +66,21 @@ class ConvergenceMonitor:
             Maximum number of iterations to perform.
         verbose : bool
             Whether per-iteration convergence reports are printed.
+        strict : bool
+            Whether to enforce that the values reported are strictly
+            increasing.
         """
         self.tol = tol
         self.n_iter = n_iter
         self.verbose = verbose
+        self.strict = strict
         self.history = deque()
         self.iter = 0
 
     def __repr__(self):
         class_name = self.__class__.__name__
         params = sorted(dict(vars(self), history=list(self.history)).items())
-        return (f"{class_name}(\n"
+        return ("{}(\n".format(class_name)
                 + "".join(map("    {}={},\n".format, *zip(*params)))
                 + ")")
 
@@ -105,6 +109,9 @@ class ConvergenceMonitor:
             message = self._template.format(
                 iter=self.iter + 1, log_prob=log_prob, delta=delta)
             print(message, file=sys.stderr)
+        if self.strict and self.history and log_prob < self.history[-1]:
+            raise ValueError(f"Model is not converging.  Current: {log_prob}"
+                             f" is not greater than {self.history[-1]}.")
         self.history.append(log_prob)
         self.iter += 1
 
@@ -117,49 +124,19 @@ class ConvergenceMonitor:
                  self.history[-1] - self.history[-2] < self.tol))
 
 
-class BaseHMM(BaseEstimator):
+class _AbstractHMM(BaseEstimator):
     """
-    Base class for Hidden Markov Models.
-
-    This class allows for easy evaluation of, sampling from, and maximum a
-    posteriori estimation of the parameters of a HMM.
-
-    Attributes
-    ----------
-    monitor_ : ConvergenceMonitor
-        Monitor object used to check the convergence of EM.
-    startprob_ : array, shape (n_components, )
-        Initial state occupation distribution.
-    transmat_ : array, shape (n_components, n_components)
-        Matrix of transition probabilities between states.
-
-    Notes
-    -----
-    Normally, one should use a subclass of `.BaseHMM`, with its specialization
-    towards a given emission model.  In rare cases, the base class can also be
-    useful in itself, if one simply wants to generate a sequence of states
-    using `.BaseHMM.sample`.  In that case, the feature matrix will have zero
-    features.
+    Base class for Hidden Markov Models learned via Expectation-Maximization
+    and Variational Bayes.
     """
 
-    def __init__(self, n_components=1,
-                 startprob_prior=1.0, transmat_prior=1.0,
-                 algorithm="viterbi", random_state=None,
-                 n_iter=10, tol=1e-2, verbose=False,
-                 params=string.ascii_letters,
-                 init_params=string.ascii_letters,
-                 implementation="log"):
+    def __init__(self, n_components, algorithm, random_state, n_iter,
+                 tol, verbose, params, init_params, implementation):
         """
         Parameters
         ----------
         n_components : int
             Number of states in the model.
-        startprob_prior : array, shape (n_components, ), optional
-            Parameters of the Dirichlet prior distribution for
-            :attr:`startprob_`.
-        transmat_prior : array, shape (n_components, n_components), optional
-            Parameters of the Dirichlet prior distribution for each row
-            of the transition probabilities :attr:`transmat_`.
         algorithm : {"viterbi", "map"}, optional
             Decoder algorithm.
         random_state: RandomState or an int seed, optional
@@ -184,28 +161,16 @@ class BaseHMM(BaseEstimator):
             to use logarithms for backwards compatability.  However, the
             scaling implementation is generally faster.
         """
+
         self.n_components = n_components
         self.params = params
         self.init_params = init_params
-        self.startprob_prior = startprob_prior
-        self.transmat_prior = transmat_prior
         self.algorithm = algorithm
-        self.random_state = random_state
         self.n_iter = n_iter
         self.tol = tol
         self.verbose = verbose
         self.implementation = implementation
-        self.monitor_ = ConvergenceMonitor(self.tol, self.n_iter, self.verbose)
-
-    def get_stationary_distribution(self):
-        """Compute the stationary distribution of states."""
-        # The stationary distribution is proportional to the left-eigenvector
-        # associated with the largest eigenvalue (i.e., 1) of the transition
-        # matrix.
-        _utils.check_is_fitted(self, "transmat_")
-        eigvals, eigvecs = linalg.eig(self.transmat_.T)
-        eigvec = np.real_if_close(eigvecs[:, np.argmax(eigvals)])
-        return eigvec / eigvec.sum()
+        self.random_state = random_state
 
     def score_samples(self, X, lengths=None):
         """
@@ -463,8 +428,8 @@ class BaseHMM(BaseEstimator):
             currstate, random_state=random_state)]
 
         for t in range(n_samples - 1):
-            currstate = (transmat_cdf[currstate] > random_state.rand()) \
-                .argmax()
+            currstate = (
+                (transmat_cdf[currstate] > random_state.rand()).argmax())
             state_sequence.append(currstate)
             X.append(self._generate_sample_from_state(
                 currstate, random_state=random_state))
@@ -494,60 +459,37 @@ class BaseHMM(BaseEstimator):
             Returns self.
         """
         X = check_array(X)
-        self._init(X)
-        self._check()
 
+        if lengths is None:
+            lengths = np.asarray([X.shape[0]])
+
+        self._init(X, lengths)
+        self._check()
         self.monitor_._reset()
 
-        impl = {
-            "scaling": self._fit_scaling,
-            "log": self._fit_log,
-        }[self.implementation]
         for iter in range(self.n_iter):
-            stats = self._initialize_sufficient_statistics()
-            curr_log_prob = 0
-            for sub_X in _utils.split_X_lengths(X, lengths):
-                lattice, log_prob, posteriors, fwdlattice, bwdlattice = \
-                        impl(sub_X)
-                # Derived HMM classes will implement the following method to
-                # update their probability distributions, so keep
-                # a single call to this method for simplicity.
-                self._accumulate_sufficient_statistics(
-                    stats, sub_X, lattice, posteriors, fwdlattice,
-                    bwdlattice)
-                curr_log_prob += log_prob
+            stats, curr_logprob = self._do_estep(X, lengths)
+
+            # Compute lower bound before updating model parameters
+            lower_bound = self._compute_lower_bound(curr_logprob)
 
             # XXX must be before convergence check, because otherwise
             #     there won't be any updates for the case ``n_iter=1``.
             self._do_mstep(stats)
-
-            self.monitor_.report(curr_log_prob)
+            self.monitor_.report(lower_bound)
             if self.monitor_.converged:
                 break
 
-        if (self.transmat_.sum(axis=1) == 0).any():
-            _log.warning("Some rows of transmat_ have zero sum because no "
-                         "transition from the state was ever observed.")
-
+            if (self.transmat_.sum(axis=1) == 0).any():
+                _log.warning("Some rows of transmat_ have zero sum because no "
+                             "transition from the state was ever observed.")
         return self
 
     def _fit_scaling(self, X):
-        frameprob = self._compute_likelihood(X)
-        log_prob, fwdlattice, scaling_factors = _hmmc.forward_scaling(
-            self.startprob_, self.transmat_, frameprob)
-        bwdlattice = _hmmc.backward_scaling(
-            self.startprob_, self.transmat_, frameprob, scaling_factors)
-        posteriors = self._compute_posteriors_scaling(fwdlattice, bwdlattice)
-        return frameprob, log_prob, posteriors, fwdlattice, bwdlattice
+        raise NotImplementedError("Must be overridden in subclass")
 
     def _fit_log(self, X):
-        log_frameprob = self._compute_log_likelihood(X)
-        log_prob, fwdlattice = _hmmc.forward_log(
-            self.startprob_, self.transmat_, log_frameprob)
-        bwdlattice = _hmmc.backward_log(
-            self.startprob_, self.transmat_, log_frameprob)
-        posteriors = self._compute_posteriors_log(fwdlattice, bwdlattice)
-        return log_frameprob, log_prob, posteriors, fwdlattice, bwdlattice
+        raise NotImplementedError("Must be overridden in subclass")
 
     def _compute_posteriors_scaling(self, fwdlattice, bwdlattice):
         posteriors = fwdlattice * bwdlattice
@@ -595,34 +537,7 @@ class BaseHMM(BaseEstimator):
         This is used to detect whether the user did not pass enough data points
         for a non-degenerate fit.
         """
-
-    def _init(self, X):
-        """
-        Initialize model parameters prior to fitting.
-
-        Parameters
-        ----------
-        X : array-like, shape (n_samples, n_features)
-            Feature matrix of individual samples.
-        """
-        self._check_and_set_n_features(X)
-        init = 1. / self.n_components
-        random_state = check_random_state(self.random_state)
-        if self._needs_init("s", "startprob_"):
-            self.startprob_ = random_state.dirichlet(
-                np.full(self.n_components, init))
-        if self._needs_init("t", "transmat_"):
-            self.transmat_ = random_state.dirichlet(
-                np.full(self.n_components, init), size=self.n_components)
-        n_fit_scalars_per_param = self._get_n_fit_scalars_per_param()
-        if n_fit_scalars_per_param is not None:
-            n_fit_scalars = sum(
-                n_fit_scalars_per_param[p] for p in self.params)
-            if X.size < n_fit_scalars:
-                _log.warning(
-                    "Fitting a model with %d free scalar parameters with only "
-                    "%d data points will result in a degenerate solution.",
-                    n_fit_scalars, X.size)
+        raise NotImplementedError("Must be overridden in subclass")
 
     def _check_sum_1(self, name):
         """Check that an array describes one or more distributions."""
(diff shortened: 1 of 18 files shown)


blckmaxima:
@alexrockhill I'd welcome your feedback too :-)

alexrockhill:
I think I'll have to read some of your thesis first :)

lib/hmmlearn/vhmm.py
```diff
@@ -0,0 +1,626 @@
...
+                if self.scale_prior is None:
+                    self.W_0_inv_ = np.broadcast_to(
+                        np.identity(self.n_features_) * 1e-3,
+                        (self.n_components, self.n_features_, self.n_features_)
+                    )
+                else:
+                    self.W_0_inv_ = self.scale_prior
+
+            elif self.covariance_type == "tied":
+                assert False, "Not Finished"
```
anntzer, line 389: `raise NotImplementedError`? ditto below

scripts/benchmark.py
```diff
@@ -17,21 +17,27 @@
 
 
 class Benchmark:
-    def __init__(self, repeat, n_iter, verbose):
+    def __init__(self, repeat, n_iter, verbose, random_state):
```
anntzer, line 20: May be worth splitting out these small independent parts into separate preliminary PRs to make the main part easier to review.

lib/hmmlearn/base.py
```diff
@@ -549,30 +558,45 @@ def _do_viterbi_pass(self, log_frameprob):
             log_frameprob)
         return log_prob, state_sequence
 
-    def _do_forward_scaling_pass(self, frameprob):
+    def _do_forward_scaling_pass(self, frameprob, startprob=None,
```
anntzer, line 561: Should these simply be transformed into private free functions to which you *always* have to pass startprob and transmat?  Ditto below.

anntzer:
Just put in some trivial comments for now (this is a pretty big PR and will take a bit of time for me to digest; of course @alexrockhill's opinion will be useful too :-))  I think it's OK to not support all covariance types and gaussian mixtures in a first step, but tests would probably be useful.

blckmaxima:
@anntzer Thanks for the feedback, I'll update things in a few days.  Slow and steady...
@alexrockhill It seems intimidating, but only Chapter 1, where I try and unify the mess of notation across all of the Variational HMM literature is needed.

lib/hmmlearn/base.py
blckmaxima: Let me try it and we can see what it looks like.

scripts/benchmark.py
blckmaxima: definitely.

alexrockhill:
Maybe it would be nice to add an example that shows the use of this new class and explains when to use it in this PR.

lib/hmmlearn/kl_divergence.py
```diff
@@ -0,0 +1,147 @@
...
+    q = np.asarray(q)
+    p = np.asarray(p)
+
+    qsum = q.sum()
+    psum = p.sum()
+    delta = q - p
+    return gammaln(qsum) - gammaln(psum) - \
+            np.sum(gammaln(q) - gammaln(p)) + \
+            np.sum(
+                np.dot(
```
anntzer, line 30: why is there both a sum and a dot?  are q and p intended to be 2D?

lib/hmmlearn/kl_divergence.py
```diff
@@ -0,0 +1,147 @@
...
+            )
+
+
+def kl_normal_distribution(mean_q, variance_q, mean_p, variance_p):
+    """
+    KL Divergence between two normal distributions
+
+    """
+    result = (np.log(variance_p / variance_q)) / 2 \
+             + (mean_q**2 + mean_p**2 + variance_q - 2 * mean_q * mean_p) \
```
anntzer, line 43: perhaps `(mean_q - mean_p) ** 2 + (the rest)` where it's more clear that the result should be positive.

lib/hmmlearn/kl_divergence.py
```diff
@@ -0,0 +1,147 @@
...
+
+def _E(dof, scale):
+    """
+
+    $L(a, B) = \int \mathcal{Wishart}(\Gamma; a, B) \log |\Gamma| d\Gamma$
+    """
+    term1 = -np.log(np.linalg.det(scale/2))
+    term2 = 0
+    for i in range(1, scale.shape[0] + 1):
+        term2 += digamma((dof + 1 - i)/2)
```
anntzer, line 133: probably `digamma((dof - np.arange(scale.shape[0])) / 2).sum()`?  ditto below.

lib/hmmlearn/vhmm.py
```diff
@@ -0,0 +1,626 @@
...
+            transmat_prior=transmat_prior,
+            algorithm=algorithm, random_state=random_state,
+            n_iter=n_iter, tol=tol, verbose=verbose,
+            params=params, init_params=init_params,
+            implementation=implementation
+        )
+
+        self.emissions_prior = emissions_prior
+
+    def _check_and_set_categorical_features(self, X):
```
anntzer, line 72: Perhaps introduce a `_BaseCategoricalHMM` that hosts the common methods (check_and_set_categorical_n_features, get_n_fit_scalars_per_param, etc.) and make both the nonvariational version and the variational one inherit from it?  This way it is easier to focus on just what's new for the variational version.

lib/hmmlearn/kl_divergence.py
blckmaxima: I'll take a few days to go over these and ping you.

lib/hmmlearn/vhmm.py
blckmaxima: If we can get the other PR merged to have a true MultinimialHMM, then I will rebase and try this.

anntzer: OK, let's try to get that other PR in first, I think it's quite close.

alexrockhill:
How similar is the motivation behind this class to a variational autoencoder? Is it basically the same concept? If so, I could maybe take a stab at an example and review in the process.

blckmaxima:
@alexrockhill I have just now added an example.  I'm only loosely familiar with variational auto-encoders, but some of the underlying math/optimization must be the same.

What might be more relevant to this project, if you  are interested in learning, is adding the Poisson emissions (EM and VI) :-)

alexrockhill:
Okay! I'll probably wait until this merges first but that would be a good thing to add.

Is the example supposed to show that the variation version has better performance? It looks like the example is simple enough that both models do pretty well and about equivalently if I'm not mistaken. Is there some situation where the non-variation version fails?

lib/hmmlearn/vhmm.py
```diff
@@ -0,0 +1,724 @@
...
+        return {
+            "s": nc - 1,
+            "t": nc * (nc - 1),
+            "m": nc * nf,
+            "c": {
+                "spherical": nc,
+                "diag": nc * nf,
+                "full": nc * nf * (nf + 1) // 2,
+                "tied": nf * (nf + 1) // 2,
+            }[self.covariance_type],
```
alexrockhill, line 460: Might be nice if this wasn't repeated from `hmm.py`

alexrockhill (commented):
From the example, looks like everything is functional and very cool. Super helpful for the future that other classes can be variational as well. It was a lot to go over but I'll take it in a bit more and hopefully do another pass before it merges.

examples/variational_inference.py
```diff
@@ -0,0 +1,115 @@
...
+    best_n, best_score = max(scores.items(), key=lambda x: x[1])
+    for n, score in scores.items():
+        flag = "* <- Best Model" if score == best_score else ""
+        print(f"{algo}({n}): {score:.4f}{flag}")
+
+    print(f"Best Model {algo}")
+    best_model = best_models[algo][best_n]
+    print(best_model.transmat_)
+    if algo == "EM":
+        print(best_model.means_)
```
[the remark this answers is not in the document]
blckmaxima: Thank you for the feedback! I will work the formatting changes next - I probably need better defaults in my editor :-)

lib/hmmlearn/base.py
```diff
@@ -856,3 +877,324 @@ def _do_mstep(self, stats):
...
+
+        Parameters
+        ----------
+        X : array-like, shape (n_samples, n_features_)
+            Feature matrix of individual samples.
+        lengths : array-like of integers, shape (n_sequences, )
+            Lengths of the individual sequences in ``X``. The sum of
+            these should be ``n_samples``.
+        """
+        init = 1. / self.n_components
```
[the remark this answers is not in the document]
alexrockhill: Nevermind, this terminology is used in `base.py` as well so maybe best to stick to it

lib/hmmlearn/vhmm.py
blckmaxima: Yes - I plan to address this, and the similar method for the categorical HMM, once the Multinomial/Categorical PR is merged.

blckmaxima:
@alexrockhill and @anntzer Thank you for all the merges and discussions the past week, and apologies for unavailable.  I will rebase this on the latest, and review the remaining tasks. One area that has been identified is seeing what common code can be shared between the EM and VI variants.

lib/hmmlearn/vhmm.py
blckmaxima: This is different enough that I think it warrants  keeping it separate.  What do you think?

lib/hmmlearn/vhmm.py
blckmaxima: Thank you for merging that other PR - I'll take a stab at this part next.

blckmaxima: @anntzer I think the way I've set things up a inheritance will be tricky as I understand it-

Here are the methods that are common:
   * `compute_likelihood`
   * `_generate_sample_from_state`
   * `_initialize_sufficient_statistics`
   * `_accumulate_sufficient_statistics`
   * ` _get_n_fit_scalars_per_param`
   * `_check_and_set_categorical_n_features` (already made free-standing)

Because of how inheritance is used, `_initialize_sufficient_statistics` and `_accumulate_sufficient_statistics`  would still need to call the parent method.  However, a `_BaseCategoricalHMM` would not derive from the necessary classes to still ensure that `super()._initialize_sufficient_statistics does the right thing.

I think we'd end up with something like this:
```mermaid
  graph TD;
      BaseEstimator-->BaseHMM
      BaseHMM-->VariationalBaseHMM;
      BaseHMM-->CategoricalHMM;
      VariationalBaseHMM-->VariationalCategoricalHMM;
      _BaseCategorial-HMM->CategoricalHMM;
      _BaseCategorial-HMM->VariationalCategoricalHMM;
```

So, I don't see a way to insert `_BaseCategoricalHMM` into the inheritance hierarchy and still make things work.  One option would be to at least make `_BaseCategoricalHMM`, with static methods, and have each of the Variational/EM classes call methods into it.  This would at least make the code centralized, and allow further evolution over time.

anntzer: Actually, super-calls should work, no?  I haven't actually looked at the details of which class implements what (so I may be misunderstanding your problem), but say you have
```
 Base
 /  \
B   C
 \ /
  D
```
and all 4 methods implement `foo` (i.e. `_initialize_sufficient_statistics`) with a super()-call (except for `Base`; this is important so that we don't end up trying calling the non-existent `object.foo()`):
```
class Base:
    def foo(self):
        print("base")
class A(Base):
    def foo(self):
        print("a"); return super().foo()
class B(Base):
    def foo(self):
        print("b"); return super().foo()
class C(A, B):
    def foo(self):
        print("c"); return super().foo()
```
then `C().foo()` will actually execute all four method bodies (printing `c a b base` -- note that the order of inheritance ends up determining whether A.foo or B.foo is called first).  (See e.g. http://python-history.blogspot.com/2010/06/method-resolution-order.html; also note that the old Py2 super-syntax, while longer, was clearer here: the result of the super-call in a method body actually depends on the exact type of `self`, not just what class we are currently in).

Or did I completely misunderstand your point?

blckmaxima: To move what I proposed, to this example, then I believe I would need `_BaseCategoricalHMM` to inherit from `BaseHMM`?,  We'll end up with two Diamond Patterns then, one for the `VariationalCategoricalHMM`, and one for the EM `CategoricalHMM`:

```mermaid
  graph TD;
      BaseEstimator-->BaseHMM;
      BaseHMM-->VariationalBaseHMM;
      BaseHMM-->CategoricalHMM;
      VariationalBaseHMM-->VariationalCategoricalHMM;
     BaseEstimator -->_BaseCategorialHMM;
      _BaseCategorialHMM-->CategoricalHMM;
      _BaseCategorialHMM-->VariationalCategoricalHMM;
```

blckmaxima: @alexrockhill @anntzer 

Making Emissions Specific classes greatly simplified things.  Right now I stuck them in `base.py`, but should we create `emissions.py` (or something similar?)  This should hopefully make the review easier.

Also, apologies for how long this has taken!

lib/hmmlearn/base.py
```diff
@@ -9,13 +9,75 @@
...
 
 
 _log = logging.getLogger(__name__)
 #: Supported decoder algorithms.
 DECODER_ALGORITHMS = frozenset(("viterbi", "map"))
 
+# Wrappers around _hmmc, used by both EM and VI
+def do_forward_scaling_pass(frameprob, startprob, transmat):
+    fwdlattice, scaling_factors = _hmmc.forward_scaling(
+        np.asarray(startprob), np.asarray(transmat),
```
anntzer, line 24: I just realized that pybind11 is perfectly happy to take array-likes and handle the casting itself (it ultimately uses PyArray_FromAny), which means that we don't need these np.asarray calls; I think it may(?) be easier to just get rid of these extra wrappers and e.g. adjust _hmmc.forward_scaling to return log_prob, fwdlattice, scaling_factors, or move log_mask_zero to the C++ side.  I can do that as a separate PR, if you prefer.

lib/hmmlearn/base.py
```diff
@@ -52,9 +114,9 @@ class ConvergenceMonitor:
     ...                                   model.monitor_.verbose)
     """
 
-    _template = "{iter:>10d} {log_prob:>16.4f} {delta:>+16.4f}"
+    _template = "{iter:>10d} {log_prob:>16.8f} {delta:>+16.8f}"
 
-    def __init__(self, tol, n_iter, verbose):
+    def __init__(self, tol, n_iter, verbose, strict=False):
```
anntzer, line 119: Let's make this keyword-only (see e.g. https://emptysqua.re/blog/api-evolution-the-right-way/).

lib/hmmlearn/base.py
```diff
@@ -858,3 +909,434 @@ def _do_mstep(self, stats):
...
+        self.n_iter = n_iter
+        self.tol = tol
+        self.verbose = verbose
+        self.params = params
+        self.init_params = init_params
+        self.implementation = implementation
+        # For the case of updating all model components
+        # at each iteration, we can be strict with the convergence
+        # monitory - we know that the Lower Bound will improve
+        # at each iteration
```
anntzer, line 936: Why not put the same strict check in the standard BaseHMM?  (Also, are you sure there's no risk of floating point slop causing non-strict improvements, as in BaseHMM?)

anntzer:
One thing a bit strange is that while VariationalBaseHMM inherits from BaseHMM, it overrides quite a few methods but doesn't call the super() methods.  Perhaps it would be cleaner to rename BaseHMM to e.g. NonVariationalBaseHMM (probably a better name is possible), and introduce a new AbstractHMM (or whatever) as common base for NonVariationalBaseHMM and VariationalBaseHMM?  This way e.g. `AbstractHMM._check` can contain the check for n_fit_scalars_per_param and both NonVariationalBaseHMM and VariationalBaseHMM can just super-call into it (ditto for other methods).

lib/hmmlearn/vhmm.py
anntzer: Sure, moving base emissions to emissions.py seems reasonable.
No worries for the time, we all have many other priorities :)

hbredin:
Not sure this is the right medium but I wanted to tell you that I have been playing a bit with this PR for speaker diarization in [pyannote.audio](https://github.com/pyannote/pyannote-audio) and would love to help make it happen. 

I am particularly interested in support for other covariance type  (diag, tied). My observations ([speaker embeddings](https://huggingface.co/pyannote/embedding)) are high-dimensional and full covariance does not work that well...

Let me know if/how I can help!

blckmaxima:
Thank you for the feedback, let me see if I can squeeze some time this next
week to get this closer to the finish line!

lib/hmmlearn/base.py
blckmaxima: Variational Algorithms guarantee that the lower bound will increase at each step.  (https://arxiv.org/abs/1601.00670)   During my thesis testing I did not see non-strict improvements due to floating point slop, though I realize that is not a strict test.

I believe in BaseHMM, the way we have priors for model parameters makes the Log likelihood not always strictly decreasing.  If the priors were all equal to zero, then we could add this check.  However, I was not confident about making this assertion there.

blckmaxima:
> One thing a bit strange is that while VariationalBaseHMM inherits from BaseHMM, it overrides quite a few methods but doesn't call the super() methods. Perhaps it would be cleaner to rename BaseHMM to e.g. NonVariationalBaseHMM (probably a better name is possible), and introduce a new AbstractHMM (or whatever) as common base for NonVariationalBaseHMM and VariationalBaseHMM? This way e.g. `AbstractHMM._check` can contain the check for n_fit_scalars_per_param and both NonVariationalBaseHMM and VariationalBaseHMM can just super-call into it (ditto for other methods).

Let me look at that next.

> 

@hbredin  - I have just add "tied" covariance.  Let me know if you have any feedback!

lib/hmmlearn/base.py
blckmaxima: Let me look at this - it will also simplify this MR.

hbredin:
Thanks for working on this! I tried fitting a model on a typical sequence of speaker embedding with no success so far (the model ends up full of NaNs).

Here is a link to a Google Colab reproducing my experiment:

https://colab.research.google.com/gist/hbredin/528ce382058273134098229d1c999ffa/variationalgaussianhmm.ipynb

lib/hmmlearn/base.py
anntzer: Sure.  Can you leave a comment to that effect in the source?

lib/hmmlearn/base.py
```diff
@@ -586,41 +553,18 @@ def _get_n_fit_scalars_per_param(self):
         for a non-degenerate fit.
         """
 
-    def _init(self, X):
+    def _init(self, X, lengths=None):
```
anntzer, line 556: You should just normalize `lengths` early (at the only call site, in fit()), and let implementations assume that it is always passed and never None.

lib/hmmlearn/base.py
```diff
@@ -482,7 +478,7 @@ def fit(self, X, lengths=None):
 
         Parameters
         ----------
-        X : array-like, shape (n_samples, n_features)
+        X : array-like, shape (n_samples, n_features_)
```
anntzer, line 481: The docs don't use a trailing underscore here (the same comment applies throughout).

anntzer:
Just had a quick look because I saw there was activity after the merge of #478, but please ping when this is ready for re-review.

blckmaxima:
@anntzer @alexrockhill  This latest version splits out HMMBase into _AbstractHMM, and removes the awkward inheritance we had before.  In doing so, I reworked fit() similar to how things are done in sklearn's Gaussian Mixture by creating e_step, an m_step, and a compute_lower_bound methods. One difference is that we have a loop over sequences, and we thus cache the subnormalized probabilities at the beginning of each e_step (see e_step_begin())
 
Given these bigger changes, please let me know if you see anything that I've missed.

lib/hmmlearn/base.py
```diff
@@ -494,60 +490,31 @@ def fit(self, X, lengths=None):
...
 
     def _fit_scaling(self, X):
-        frameprob = self._compute_likelihood(X)
-        log_prob, fwdlattice, scaling_factors = _hmmc.forward_scaling(
-            self.startprob_, self.transmat_, frameprob)
-        bwdlattice =  _hmmc.backward_scaling(
-            self.startprob_, self.transmat_, frameprob, scaling_factors)
-        posteriors = self._compute_posteriors_scaling(fwdlattice, bwdlattice)
-        return frameprob, log_prob, posteriors, fwdlattice, bwdlattice
+        """
```
anntzer, line 512: `raise NotImplementedError("Must be overridden in subclass")` rather than an empty docstring? ditto below.

lib/hmmlearn/kl_divergence.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+
+    KL Divergence between two Wishart Distributions
+
+    q(x) = Wishart(R|dof_q, scale_q)
+    p(x) = Wishart(R|dof_p, scale_p)
+
+    Definition from:
+        Shihao Ji, B. Krishnapuram, and L. Carin,
+        “Variational Bayes for continuous hidden Markov models and its
```
anntzer, line 99: Use straight quotes (below too).

lib/hmmlearn/vhmm.py
```diff
@@ -0,0 +1,611 @@
...
+                )
+            elif self.covariance_type == "spherical":
+                raise NotImplementedError(
+                    f"{self.covariance_type} is not implemented yet"
+                )
+            else:
+                raise ValueError(
+                    f"Unknown covariance_type: {self.covariance_type}")
+
+        if self._needs_init("w", "W_k_"):
```
anntzer, line 319: Are these W_foo attributes that the end user is supposed to sometimes want to set themselves?  If so they should be documented (at least cursorily), otherwise they should be private.

lib/hmmlearn/base.py
```diff
@@ -494,60 +490,31 @@ def fit(self, X, lengths=None):
...
             #     there won't be any updates for the case ``n_iter=1``.
             self._do_mstep(stats)
-
-            self.monitor_.report(curr_log_prob)
+            self.monitor_.report(lower_bound)
             if self.monitor_.converged:
                 break
-
-        if (self.transmat_.sum(axis=1) == 0).any():
-            _log.warning("Some rows of transmat_ have zero sum because no "
```
anntzer, line 529: I think this check/warning got lost?

lib/hmmlearn/vhmm.py
blckmaxima: Work in progress, and will actually remove a set of `W` for simplicity.

blckmaxima: I will work on this next.

lib/hmmlearn/base.py
blckmaxima: Should be cleaned up now - we've since simplified this

lib/hmmlearn/kl_divergence.py
blckmaxima: Cleaned up using einsum.

lib/hmmlearn/vhmm.py
blckmaxima: We've since restructured everything.

lib/hmmlearn/base.py
```diff
@@ -66,17 +98,21 @@ def __init__(self, tol, n_iter, verbose):
             Maximum number of iterations to perform.
         verbose : bool
             Whether per-iteration convergence reports are printed.
+        strict : bool
+            Whether to enforce that the values reported are strictly
+            increasing
```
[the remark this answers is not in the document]
blckmaxima: Done.

lib/hmmlearn/base.py
blckmaxima: Done.

lib/hmmlearn/base.py
blckmaxima: Done.

lib/hmmlearn/base.py
blckmaxima: Done in fit() now.

lib/hmmlearn/base.py
blckmaxima: done.

lib/hmmlearn/vhmm.py
blckmaxima: I've renamed these to match what I found in a reference `W-> scale`

blckmaxima:
@anntzer @alexrockhill  It's been awhile!  As you are able, would you mind reviewing this MR?

Major changes:
* Incorporating the structural changes that were recommended
* Finished the Gaussian implementation for the four different Covariance Types.

If you have recommendations(or tools) for improving style please send them my way.  I mostly let `vim` do what it wants to keep within the line length limit.

anntzer:
This looks OK to me but the PR is enormous and it is really difficult to follow all the moving parts.
I've tried to split out some parts that can reasonably considered as preliminary work at https://github.com/anntzer/hmmlearn/tree/vhmm (specifically, consistently moving out all emissions to _emissions.py (\*), making _check_and_set_n_features a method on the class which allows overriding it more easily, and reorder the emissions in a consistent (actually, alphabetical) order (Categorical, Gaussians, Multinomial, Poisson)); can you try to rebase the rest of the PR on top of that?
I have minor style nitpicks but I can fix that after.

(\*) it seems easier to just make the whole module private rather than have a public module with no public classes in it.

blckmaxima:
@anntzer Thank you so much for your time, and work.  Yes, it is enormous due to how I took your suggestions to refactor some of the classes. In the end, it will be worth it, but as of now, it is daunting.

I'll rebase and see how it goes!

anntzer:
Thanks for doing the rebase.  I allowed myself some more tweaking plus a separate style cleanup commit (force-pushed to your repo); if that looks good to you I'll merge :)

blckmaxima:
Thanks - I might disagree about some of your style choices, but appreciate your overall touch.  Merge away!

anntzer:
Thanks for all the work, and happy new year :)

anntzer:
Can you write a changelog entry, by the way? Seems like something worth advertising a bit.

alexrockhill:
Hey sorry I didn't get to review this again, looks good. I went to check the example on the documentation for latest and the example related to this had a link that didn't work. Not sure if that's related to the failed CI on the last merge or what but that would be nice to fix.

anntzer:
Thanks, good catch, this should now be fixed.

PR #465, opened 2022-05-14
[ENH, MRG] Add emissionprob_prior

alexrockhill:
This implements a to do, allowing the user to pass an `emissionprob_prior` for consistency.

### DIFF
--- a/lib/hmmlearn/hmm.py
+++ b/lib/hmmlearn/hmm.py
@@ -360,8 +360,7 @@ class MultinomialHMM(BaseHMM):
     MultinomialHMM(algorithm='viterbi',...
     """
 
-    # TODO: accept the prior on emissionprob_ for consistency.
-    def __init__(self, n_components=1,
+    def __init__(self, n_components=1, emissionprob_prior=None,
                  startprob_prior=1.0, transmat_prior=1.0,
                  algorithm="viterbi", random_state=None,
                  n_iter=10, tol=1e-2, verbose=False,
@@ -372,6 +371,9 @@ class MultinomialHMM(BaseHMM):
         ----------
         n_components : int
             Number of states.
+
+        emissionprob_prior : array, shape (n_components, n_features), optional
+            The prior on the emission parameters for each state.
 
         startprob_prior : array, shape (n_components, ), optional
             Parameters of the Dirichlet prior distribution for
@@ -418,6 +420,7 @@ class MultinomialHMM(BaseHMM):
                          n_iter=n_iter, tol=tol, verbose=verbose,
                          params=params, init_params=init_params,
                          implementation=implementation)
+        self.emissionprob_prior = emissionprob_prior
 
     score_samples, score, decode, predict, predict_proba, sample, fit = map(
         _multinomialhmm_fix_docstring_shape, [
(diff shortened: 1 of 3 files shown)


lib/hmmlearn/hmm.py
```diff
@@ -359,8 +359,7 @@ class MultinomialHMM(BaseHMM):
     MultinomialHMM(algorithm='viterbi',...
     """
 
-    # TODO: accept the prior on emissionprob_ for consistency.
-    def __init__(self, n_components=1,
+    def __init__(self, n_components=1, emissionprob_prior=1.0,
```
anntzer, line 362: This means that anyone who previously passed more than one positional argument will now accidentally set emissionprob_prior instead of startprob_prior.  While such a use of the API seems quite unlikely, you can avoid anyone silently getting wrong results by making all parameters except for n_components keyword-only (with a `*`) -- now at least anyone with the old usage will have an exception and will know they need to update their code.

Also, for consistency with GaussianHMM, emissionprob_prior should come after startprob_prior and transmat_prior.

Other than that, this looks good to me.

anntzer (changes requested):

lib/hmmlearn/hmm.py
alexrockhill: Good point, I did think about that but I didn't have the solution, thanks for that!

anntzer (approved):
I'll give a couple of days to @blckmaxima in case he wants to comment.

alexrockhill:
This one is pretty minor and has no effect unless the prior is specified by an argument, might be okay to merge unless you want to wait longer for comments.

lib/hmmlearn/hmm.py
```diff
@@ -485,8 +489,9 @@ def _accumulate_sufficient_statistics(self, stats, X, lattice,
     def _do_mstep(self, stats):
         super()._do_mstep(stats)
         if 'e' in self.params:
-            self.emissionprob_ = (
-                stats['obs'] / stats['obs'].sum(axis=1, keepdims=True))
+            self.emissionprob_ = np.maximum(
+                self.emissionprob_prior - 1 + stats['obs'],
+                0) / stats['obs'].sum(axis=1, keepdims=True)
```
anntzer, line 494: Actually, doesn't the normalization need to be fixed here?

alexrockhill: Ah yeah, good catch

lib/hmmlearn/hmm.py
```diff
@@ -359,9 +359,8 @@ class MultinomialHMM(BaseHMM):
     MultinomialHMM(algorithm='viterbi',...
     """
 
-    # TODO: accept the prior on emissionprob_ for consistency.
-    def __init__(self, n_components=1,
-                 startprob_prior=1.0, transmat_prior=1.0,
+    def __init__(self, n_components=1, startprob_prior=1.0,
+                 transmat_prior=1.0, *, emissionprob_prior=1.0,
```
anntzer, line 363: This also needs a changelog entry ("For MultinomialHMM, parameters after transmat_prior are now keyword-only.")

anntzer (approved):

PR #466, opened 2022-05-14
[ENH, MRG] Add Dirichlet random initialization of priors on startprob and transmat

alexrockhill:
Fixes https://github.com/hmmlearn/hmmlearn/issues/419

PR #467, opened 2022-05-14
[ENH, MRG] Add Dirichlet random priors in example

alexrockhill:
This follows along with the recommendations of https://github.com/hmmlearn/hmmlearn/pull/466, using uniform random from 0 to 0.25 is less principled I think.

### DIFF
--- a/examples/plot_casino_example.py
+++ b/examples/plot_casino_example.py
@@ -33,7 +33,7 @@ from hmmlearn import hmm
 gen_model = hmm.MultinomialHMM(n_components=2, random_state=99)
 
 # the first state is the fair die so let's start there so no one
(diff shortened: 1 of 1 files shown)


anntzer (approved):
Seems reasonable to me?

examples/plot_casino_example.py
```diff
@@ -87,24 +87,22 @@
 
 best_score = best_model = None
 n_fits = 50
-np.random.seed(13)
+rng = np.random.default_rng(13)
```
anntzer, line 90: We don't require a version of numpy that necessarily has the new RNG API, and I would rather not bump the requirement just for that, so let's stick to the old RNG API.

PR #468, opened 2022-05-15
[ENH, WIP] Add PoissonHMM

alexrockhill:
Adds a `PoissonHMM` with an example.

I think this is somewhat close so if you have time, a review would be great @anntzer @blckmaxima.

I'm not sure if there is a standard we could compare to like the Wikipedia one for the `MultiNomialHMM` or if that's necessary.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ Version 0.2.7
 
 Released on February 10th, 2022.
 
+- The ``PoissonHMM`` class was added with an example use case.
 - Dropped support for Py3.5 (due to the absence of manylinux wheel supporting
   both Py3.5 and Py3.10).
 - ``_BaseHMM`` has been promoted to public API and has been renamed to
--- a/examples/plot_poisson_hmm.py
+++ b/examples/plot_poisson_hmm.py
@@ -1,11 +1,11 @@
 """
-Sampling from and decoding an HMM
----------------------------------
+Sampling from and decoding a Poisson HMM
+----------------------------------------
 
-This script shows how to sample points from a Hidden Markov Model (HMM):
-we use a 4-state model with specified mean and covariance.
+This script shows how to sample points from a Poisson Hidden Markov Model
+(HMM) with three states:
 
-The plot shows the sequence of observations generated with the transitions
+The plots show the sequence of observations generated with the transitions
 between them. We can see that, as specified by our transition matrix,
 there are no transition between component 1 and 3.
 
@@ -17,32 +17,26 @@ import matplotlib.pyplot as plt
 
 from hmmlearn import hmm
 
-# Prepare parameters for a 4-components HMM
+# Prepare parameters for a 3-components HMM
 # Initial population probability
-startprob = np.array([0.6, 0.3, 0.1, 0.0])
-# The transition matrix, note that there are no transitions possible
-# between component 1 and 3
-transmat = np.array([[0.7, 0.2, 0.0, 0.1],
-                     [0.3, 0.5, 0.2, 0.0],
-                     [0.0, 0.3, 0.5, 0.2],
-                     [0.2, 0.0, 0.2, 0.6]])
+startprob = np.array([0.6, 0.3, 0.1])
+# The transition matrix
+transmat = np.array([[0.1, 0.2, 0.7],
+                     [0.3, 0.5, 0.2],
+                     [0.5, 0.5, 0.0]])
 # The means of each component
-means = np.array([[0.0, 0.0],
-                  [0.0, 11.0],
-                  [9.0, 10.0],
-                  [11.0, -1.0]])
-# The covariance of each component
-covars = .5 * np.tile(np.identity(2), (4, 1, 1))
+lambdas = np.array([[17.4, 22.1],
+                    [35.3, 60.8],
+                    [50.1, 12.9]])
 
 # Build an HMM instance and set parameters
-gen_model = hmm.GaussianHMM(n_components=4, covariance_type="full")
+gen_model = hmm.PoissonHMM(n_components=3, random_state=99)
 
 # Instead of fitting it from the data, we directly set the estimated
 # parameters, the means and covariance of the components
 gen_model.startprob_ = startprob
 gen_model.transmat_ = transmat
-gen_model.means_ = means
-gen_model.covars_ = covars
+gen_model.lambdas_ = lambdas
 
 # Generate samples
 X, Z = gen_model.sample(500)
@@ -50,10 +44,10 @@ X, Z = gen_model.sample(500)
 # Plot the sampled data
 fig, ax = plt.subplots()
 ax.plot(X[:, 0], X[:, 1], ".-", label="observations", ms=6,
-        mfc="orange", alpha=0.7)
+        mfc="orange", alpha=0.7, linewidth=0.1)
 
 # Indicate the component numbers
-for i, m in enumerate(means):
+for i, m in enumerate(lambdas):
     ax.text(m[0], m[1], 'Component %i' % (i + 1),
             size=17, horizontalalignment='center',
             bbox=dict(alpha=.7, facecolor='w'))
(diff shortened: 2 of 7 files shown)


blckmaxima:
For an example, I like the earthquakes data from Zucchini, MacDonald, and Langrock (Hiddne Markov Models for Time Series). It comes from the US Geological Survey. 

References: https://ayorho.files.wordpress.com/2011/05/chapter1.pdf
and in a csv here: https://gitlab.com/mdanielson/hmms-public/-/blob/master/hmm/datasets/data/earthquakes.csv

alexrockhill:
Thanks @blckmaxima! It looks like this implementation was able to model the earthquake data very similarly to the one in the textbook, which gives me a lot more confidence that I did it right 😄

CHANGES.rst
```diff
@@ -8,6 +8,7 @@ Version 0.2.7
 
 Released on February 10th, 2022.
 
+- The ``PoissonHMM`` class was added with an example use case.
```
anntzer, line 11: This should go to a new
```rst
next
----
```
section, as this won't be in 0.2.7.

examples/plot_poisson_hmm.py
```diff
@@ -0,0 +1,114 @@
...
+bins = sorted(np.unique(earthquakes))
+
+fig, ax = plt.subplots()
+ax.hist(earthquakes, bins=bins, density=True)
+ax.plot(bins, np.dot(poisson.pmf(bins, model.lambdas_).T,
+                     prop_per_state[:, None]))
+ax.set_title('Histogram of Earthquakes with Fitted Poisson States')
+ax.set_xlabel('Number of Earthquakes')
+ax.set_ylabel('Proportion')
+fig.show()
```
anntzer, line 114: You can drop the fig.show(), and you need a plt.show() at the end so that the figures don't get hidden if running straight from the terminal (`python examples/plot_poisson_hmm.py`).

lib/hmmlearn/hmm.py
```diff
@@ -1049,3 +1050,148 @@ def outer_f(x):  # Outer product over features.
...
+        verbose : bool, optional
+            Whether per-iteration convergence reports are printed to
+            :data:`sys.stderr`.  Convergence can also be diagnosed using the
+            :attr:`monitor_` attribute.
+
+        params, init_params : string, optional
+            The parameters that get updated during (``params``) or initialized
+            before (``init_params``) the training.  Can contain any
+            combination of 's' for startprob, 't' for transmat, and 'e' for
+            emissionprob.  Defaults to all parameters.
```
anntzer, line 1122: Needs fix ("e" -> "l", etc.)

lib/hmmlearn/hmm.py
```diff
@@ -1049,3 +1050,148 @@ def outer_f(x):  # Outer product over features.
...
+
+    transmat_ : array, shape (n_components, n_components)
+        Matrix of transition probabilities between states.
+
+    lambdas_ : array, shape (n_components, n_features)
+        The expectation value of the waiting time parameters for each
+        feature in a given state.
+    """
+
+    def __init__(self, n_components=1, lambda_prior=1.0,
```
anntzer, line 1075: should be `lambdas_prior` (plural), consistently with `means_prior`, for example.

lib/hmmlearn/hmm.py
```diff
@@ -1049,3 +1050,148 @@ def outer_f(x):  # Outer product over features.
...
+                         implementation=implementation)
+        self.lambda_prior = lambda_prior
+
+    def _init(self, X):
+        _check_and_set_n_features(self, X)
+        super()._init(X)
+        self.random_state = check_random_state(self.random_state)
+
+        if self._needs_init('l', 'lambdas_'):
+            self.lambdas_ = self.random_state.gamma(
```
anntzer, line 1145: Is there a reason for a gamma distribution here?

anntzer:
A bunch of minor points, but mostly looks good.  I don't really have the bandwidth to review the exact math, but I'll trust you on that.

lib/hmmlearn/hmm.py
alexrockhill: It just seemed like a reasonable prior for Poisson `lambda` parameters. I'm not sure what other people have used. It seems to work fairly well for a somewhat broad range of data, potentially if `lamdas` are very large, it could be a poor choice but I think that's fairly rare and it's just the starting point.

anntzer: Would it be perhaps slightly more robust to scale the initial guess e.g. to the inverse of the mean of the observations?

alexrockhill:
> A bunch of minor points, but mostly looks good. I don't really have the bandwidth to review the exact math, but I'll trust you on that.

I checked on the example encoding and decoding data and it seemed reasonable but I was going to use it for a project sometime soon so I'll really know then.

anntzer:
Sure; the other points should still be addressed, though.

alexrockhill:
Definitely, I'm fixing the points from your review now.

lib/hmmlearn/hmm.py
alexrockhill: That's a great idea! It does look like gamma is the conjugate prior for poisson distribution so I think this is the best option, I just forgot why :) https://vioshyvo.github.io/Bayesian_inference/conjugate-distributions.html

examples/plot_poisson_hmm.py
```diff
@@ -0,0 +1,114 @@
+"""
+Using a Hidden Markov Model with Poisson Emissions to Understand Earthquakes
+----------------------------------------------------------------------------
+
+Let's look at data of magnitude 7+ earthquakes between 1900-2006 in the
+world collected by the US Geological Survey as described in this textbook:
+https://ayorho.files.wordpress.com/2011/05/chapter1.pdf. The goal is to
```
blckmaxima, line 7: Minor Nit - let's include the title of the text book with authors.

blckmaxima:
Math looks like what I've done before :+1:   I previously had initialized the lambdas with Gaussian Mixture, glad you found the Gamma link.  Looks good.

alexrockhill:
Huh so there was a bug where `stats['nobs']` wasn't being updated even though I called `super._accumulate_sufficient_statistics`. I changed it so that it worked but for the life of me, I'm really not sure why it needed to be changed, I thought the `stats` dict was mutable and therefore changed when you did `stats['nobs'] += 1`. Anyway, as long as the `nobs` gets updated properly, this looks better with adding the prior to be consistent with the other models. The results match the textbook fairly well too although the log probabilities are higher (less negative).

alexrockhill:
Huh, this passes locally for me so not sure what's causing it not to converge on some random trials. Maybe it's just not my day for bugs and I should come back to it

alexrockhill:
Ok so it was just an issue with a bad random seed for the gamma which made it regress because the prior was more accurate than the early posteriors.

I think the implementation for the gamma is correct now but if you want to check @blckmaxima, that would be more than welcome. The results look pretty reasonable but I think it suffers from the same over-fitting problem so it might be nice to have a variational version in due time.

lib/hmmlearn/hmm.py
```diff
@@ -1049,3 +1050,170 @@ def outer_f(x):  # Outer product over features.
...
+                         implementation=implementation)
+        self.lambdas_prior = lambdas_prior
+        self.lambdas_weight = lambdas_weight
+
+    def _init(self, X):
+        _check_and_set_n_features(self, X)
+        super()._init(X)
+        self.random_state = check_random_state(self.random_state)
+
+        exp_X = X.mean()
```
anntzer, line 1146: I guess this means "expectation of X", but can easily be incorrectly read as "exponential(X)". `mean_X` seems just fine?

anntzer (approved):
Seems reasonable, just a minor nit.  Still trusting you on the math :-)

If your question about `stats` is wrt when you changed the 'obs' update to `stats['nobs'] = iter + 1` in Base.fit, this certainly seemed wrong?  That's certainly not the number of observations anymore...

alexrockhill:
Yes, I got mixed up between the splitting and the number of samples in X. It should be fixed now but feel free to double check if you have time or maybe @blckmaxima can give it a quick look over. The one thing I'm not totally certain of is the prior implementation because it has the number of samples whereas the inverse gamma in GMMHMM does not. I was following the referenced textbook but might be nice to have a quick look over before merging if anyone has a minute.

blckmaxima:
I'll have time tomorrow morning to review this.

blckmaxima (approved):
To the best of my analysis, the math matches up with what I've done previously.  I had previously seen poisson mixture models initialized with a GMM, but the Gamma initialization looks like it works fine here.

anntzer:
@alexrockhill do you want to rename exp_X to mean_X? Feel free to self merge with or without this change.

Issue #469, opened 2022-05-17
[ENH] Add a uniform prior option


alexrockhill:
Related to https://github.com/hmmlearn/hmmlearn/pull/468. We enforce a gamma prior based on the method of moments fit to the parameters from the training data (`X`). This is consistent with the other HMMs which choose the best prior for you. In the Poission case, it might make sense to have a uniform prior as an option for faster convergence when one or more of the states has a waiting time very different from the mean waiting time of the distribution. This would be a very different API though with something like `lambdas_prior_dist='gamma'` by default with the option of setting to `uniform`. The parameters for the prior would then have to be passed in using `lambdas_prior` and `lambdas_weight` where if they were extra as in the uniform case, they would be ignored. Another option would be to allow a boolean `uniform_prior` keyword argument that would use the uniform prior. That's more backward compatible but less extensible. Interested to know what you think @anntzer and @blckmaxima. Maybe this is just a theoretical issue but I did have to increase `n_iters` for the PoissonHMM example to get convergence.

anntzer:
I don't claim any expertise here (usually I just retry a couple of times to see if I can get a better convergence...), but (unless there are strong principled reasons why two different init methods should be considered) I would just stick to one initialization method and let the end user manually set a starting point (and switch off init_params) if they want to do it in another way.  IOW I don't think it's really worth complexifying the API for that(?)

alexrockhill:
I think the main issue is that there are no arguments that you can pass to avoid the prior, it's coded unavoidably here for example:

```
if 'l' in self.params:
    # Based on: Hyvönen & Tolonen, "Bayesian Inference 2019"
    # section 3.2
    # https://vioshyvo.github.io/Bayesian_inference
    alphas, betas = self.lambdas_prior, self.lambdas_weight
    n = stats['nsamples']
    kappas = betas / (betas + n)
    y_bar = stats['obs'] / stats['post'][:, None]
    self.lambdas_ = kappas * (alphas / betas) + (1 - kappas) * y_bar
```

You can pass a different `lambdas_prior` and `lamdas_weight` but it's still enforcing a gamma distribution prior. This is the case for all the HMMs also.

alexrockhill:
Here are the inverse gamma and inverse Wishart distributions for the `GaussianHMM` covariance for another example.

```
if 'c' in self.params:
    covars_prior = self.covars_prior
    covars_weight = self.covars_weight
    meandiff = self.means_ - means_prior

[...]

PR #470, opened 2022-05-18
[MAINT, MRG] Fix prior for PoissonHMM

alexrockhill:
Fixes https://github.com/hmmlearn/hmmlearn/issues/469. Cleans up https://github.com/hmmlearn/hmmlearn/pull/468.

This test script from a prior commit is a good sanity check too.

```
import numpy as np
import matplotlib.pyplot as plt

from hmmlearn import hmm

# Prepare parameters for a 3-components HMM
# Initial population probability
startprob = np.array([0.6, 0.3, 0.1])
# The transition matrix
transmat = np.array([[0.1, 0.2, 0.7],
                     [0.3, 0.5, 0.2],
                     [0.5, 0.5, 0.0]])
# The means of each component
lambdas = np.array([[17.4, 22.1],
                    [35.3, 60.8],
                    [50.1, 12.9]])

# Build an HMM instance and set parameters
gen_model = hmm.PoissonHMM(n_components=3, random_state=99)

# Instead of fitting it from the data, we directly set the estimated
# parameters, the means and covariance of the components
gen_model.startprob_ = startprob
gen_model.transmat_ = transmat
gen_model.lambdas_ = lambdas

# Generate samples
X, Z = gen_model.sample(500)

# Plot the sampled data
fig, ax = plt.subplots()
ax.plot(X[:, 0], X[:, 1], ".-", label="observations", ms=6,
        mfc="orange", alpha=0.7, linewidth=0.1)

# Indicate the component numbers
for i, m in enumerate(lambdas):
    ax.text(m[0], m[1], 'Component %i' % (i + 1),
            size=17, horizontalalignment='center',
            bbox=dict(alpha=.7, facecolor='w'))
ax.legend(loc='best')
fig.show()

# %%
# Now, let's ensure we can recover our parameters.

scores = list()
models = list()
for n_components in range(1, 5):
    for idx in range(10):   # different random seeds
        # define our hidden Markov model
        model = hmm.PoissonHMM(n_components=3, random_state=idx)
        model.fit(X[:X.shape[0] // 2])  # 50/50 train/validate
        models.append(model)
        scores.append(model.score(X[X.shape[0] // 2:]))
        print(f'Converged: {model.monitor_.converged}'
              f'\tScore: {scores[-1]}')

# get the best model
model = models[np.argmax(scores)]
print(f'The best model had a score of {max(scores)}')

# use the Viterbi algorithm to predict the most likely sequence of states
# given the model
states = model.predict(X)

# %%
# Let's plot our states compared to those generated and our transition matrix
# to get a sense of our model. We can see that the recovered states follow
# the same path as the generated states, just with the identities of the
# states transposed (i.e. instead of following a square as in the first
# figure, the nodes are switch around but this does not change the basic
# pattern). The same is true for the transition matrix.

# plot model states over time
fig, ax = plt.subplots()
ax.plot(Z, states)
ax.set_title('States compared to generated')
ax.set_xlabel('Generated State')
ax.set_ylabel('Recovered State')
fig.show()

# plot the transition matrix
fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(8, 5))
ax1.imshow(gen_model.transmat_, aspect='auto', cmap='spring')
ax1.set_title('Generated Transition Matrix')
ax2.imshow(model.transmat_, aspect='auto', cmap='spring')
ax2.set_title('Recovered Transition Matrix')
for ax in (ax1, ax2):
    ax.set_xlabel('State To')
    ax.set_ylabel('State From')

fig.tight_layout()
fig.show()
```

I also realized you could get `n` from the sum of the posterior since each row adds up to 1 and there are `n_samples` number of columns so that's nice to simplify a bit.

I feel much better about this implementation. If you pass `alpha=0` and `beta=0` you'll just get `obs` / `post` which I think is important to be able to not enforce a gamma-distributed prior.

### DIFF
--- a/lib/hmmlearn/hmm.py
+++ b/lib/hmmlearn/hmm.py
@@ -1211,10 +1211,11 @@ class PoissonHMM(BaseHMM):
             # section 3.2
             # https://vioshyvo.github.io/Bayesian_inference
             alphas, betas = self.lambdas_prior, self.lambdas_weight
+            n = stats['post'].sum()
+            y_bar = stats['obs'] / stats['post'][:, None]
             # the same as kappa notation (more intuitive) but avoids divide by
             # 0, where:
             # kappas = betas / (betas + stats['post'][:, None])
             # self.lambdas_ = \
             #     kappas * (alphas / betas) + (1 - kappas) * stats['obs']
-            self.lambdas_ = \
-                (alphas + stats['obs']) / (betas + stats['post'][:, None])
+            self.lambdas_ = (alphas + n * y_bar) / (betas + n)
--- a/lib/hmmlearn/hmm.py
+++ b/lib/hmmlearn/hmm.py
@@ -1191,7 +1191,6 @@ class PoissonHMM(BaseHMM):
 
     def _initialize_sufficient_statistics(self):
         stats = super()._initialize_sufficient_statistics()
-        stats['nsamples'] = 0
         stats['post'] = np.zeros(self.n_components)
         stats['obs'] = np.zeros((self.n_components, self.n_features))
         return stats
@@ -1201,7 +1200,6 @@ class PoissonHMM(BaseHMM):
         super()._accumulate_sufficient_statistics(
             stats, obs, lattice, posteriors, fwdlattice, bwdlattice)
         if 'l' in self.params:
-            stats['nsamples'] += obs.shape[0]
             stats['post'] += posteriors.sum(axis=0)
             stats['obs'] += np.dot(posteriors.T, obs)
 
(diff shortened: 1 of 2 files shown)


lib/hmmlearn/hmm.py
```diff
@@ -1213,7 +1211,10 @@ def _do_mstep(self, stats):
             # section 3.2
             # https://vioshyvo.github.io/Bayesian_inference
             alphas, betas = self.lambdas_prior, self.lambdas_weight
-            n = stats['nsamples']
-            kappas = betas / (betas + n)
+            n = stats['post'].sum()
```
anntzer, line 1214: Actually, that's just `stats["nobs"]`, I think?

anntzer: Ah, I see.  That's a bit confusingly named, and moreover, never used afaict; probably we should just get rid of it at some point.

anntzer (approved):

lib/hmmlearn/hmm.py
alexrockhill: Agreed and quite, I originally tried to use it and was pretty confused

alexrockhill:
Ok, I think I'll merge this unless there are any more suggestions

anntzer:
This seems safe enough to go in as is.

anntzer:
@alexrockhill Would you agree to change the default lambdas_prior/lambdas_weight to be both zero?  I would prefer a uniform prior as default, for consistency with the rest (notwithstanding the very slight regularization present in GaussianHMM/GMMHMM).

alexrockhill:
Sure, sounds very reasonable.

anntzer:
Do you want to write the patch? otherwise I can do it too.

PR #471, opened 2022-05-18
_generate_sample_from_state can assume that random_state is normalized.

anntzer:
`sample` is the only public API that calls this method and it already
does `random_state = check_random_state(random_state)`.


anntzer:
Self-merging as hopefully uncontroversial.

PR #473, opened 2022-05-19
[MAINT, MRG] Add uniform prior to Poisson

### LINKED ISSUE #469: [ENH] Add a uniform prior option

alexrockhill:
Fixes #469

### DIFF
--- a/lib/hmmlearn/hmm.py
+++ b/lib/hmmlearn/hmm.py
@@ -1073,8 +1073,8 @@ class PoissonHMM(BaseHMM):
     """
 
     def __init__(self, n_components=1, startprob_prior=1.0,
-                 transmat_prior=1.0, lambdas_prior=None,
-                 lambdas_weight=None,
+                 transmat_prior=1.0, lambdas_prior=0.0,
+                 lambdas_weight=0.0,
                  algorithm="viterbi", random_state=None,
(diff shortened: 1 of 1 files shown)


alexrockhill:
Sorry, my histories got all messed up because I tried pushing to my `prior` branch again. This one works now.

anntzer (approved):

PR #478, opened 2022-06-26
Begin to simplify python/C++ interface.

blckmaxima:


blckmaxima:
@anntzer Is this approximately what you had in mind ?

lib/hmmlearn/base.py
```diff
@@ -544,36 +544,27 @@ def _fit_log(self, X):
         return log_frameprob, log_prob, posteriors, fwdlattice, bwdlattice
 
     def _do_viterbi_pass(self, log_frameprob):
-        state_sequence, log_prob = _hmmc.viterbi(
-            log_mask_zero(self.startprob_), log_mask_zero(self.transmat_),
+        return _hmmc.viterbi(self.startprob_, self.transmat_,
```
anntzer, line 547: I would even just get rid of these `_do_foo_pass` wrappers and directly inline the calls to the C hmmc functions at the call sites.

src/_hmmc.cpp
```diff
@@ -26,7 +27,33 @@ double logsumexp(double const* v, ssize_t n)
   return std::log(acc) + max;
 }
 
-std::pair<py::array_t<double>, py::array_t<double>> forward_scaling(
+void log_mask_zero_1d(py::array_t<double> probs_,
```
anntzer, line 30: You could just return a `py::array_t<double>` here (allocating it from inside the function) instead of allocating it outside and modifying it in-place.

src/_hmmc.cpp
```diff
@@ -26,7 +27,33 @@ double logsumexp(double const* v, ssize_t n)
...
+  py::array_t<double> log_probs_){
+  auto probs = probs_.unchecked<1>();
+  auto nc = probs.shape(0);
+  auto log_probs = log_probs_.mutable_unchecked<1>();
+
+  for(auto i = 0; i < nc; ++i){
+     log_probs(i) = std::log(probs(i));
+  }
+
```
anntzer, line 39: I guess that to be perfectly clean you need to clear the C-level floating point exception flag here and below (if any has been set by a zero probs).  You can probably just inline the code at https://github.com/numpy/numpy/blob/54c2c1ebd290c2cc2faa6f6272f4b04fd0c1eea5/numpy/core/src/npymath/ieee754.c.src#L619

anntzer:
Yes, just some minor nits.

src/_hmmc.cpp
blckmaxima: So I had this originally, and then saw a strange issue with memory being reused when computing the `log_xi_sums` below.  However, not doing something like:
`auto log_transmat = log_mask_zero_2d(transmat).unchecked<2>()` and instead doing:

`auto log_transmat_ = log_mask_zero_2d(transmat);
auto log_transmat = log_transmat_.unchecked<2>()`
worked.

lib/hmmlearn/base.py
blckmaxima: Done.

src/_hmmc.cpp
blckmaxima: Is it enough to test/clear `FE_DIVBYZERO` or do we want to do all of them that numpy did?

anntzer: Let's just clear it for now and open an issue that we may want to copy numpy's entire thing in the future.

blckmaxima: done.

blckmaxima:
That would be great, thank you.

blckmaxima:
This looks great - my last foray into c++/python interoperability is circa ~2007 - this is so much nicer to work with.

PR #489, opened 2022-10-25
build for python 3.11

njzjz:
Python 3.11 is released today.


anntzer:
Thanks for the note, but as you can notice python 3.11 is not available on GitHub actions yet.

PR #498, opened 2023-01-01
Finish the Gaussian vhmm

blckmaxima:
Round out tests for Guassian vhmm

### DIFF
--- a/lib/hmmlearn/tests/test_variational_gaussian.py
+++ b/lib/hmmlearn/tests/test_variational_gaussian.py
@@ -4,7 +4,8 @@ from sklearn.utils import check_random_state
 
 from hmmlearn import hmm, vhmm
 from . import (
-    assert_log_likelihood_increasing, compare_variational_and_em_models)
+    assert_log_likelihood_increasing, compare_variational_and_em_models,
+    make_covar_matrix, normalized)
 
 
 def get_mcgrory_titterington():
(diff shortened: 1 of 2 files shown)


lib/hmmlearn/tests/test_variational_gaussian.py
```diff
@@ -265,33 +285,313 @@ def test_incorrect_init_covariance(self, implementation):
...
+    def test_initialization(self, implementation):
         random_state = check_random_state(234234)
-        sequences, lengths = get_sequences(500, 1,
-                                           model=get_mcgrory_titterington())
+        sequences, lengths = get_sequences(
+            50, 10, model=get_mcgrory_titterington())
+
+        # dof's have wrong shape
+        with pytest.raises(ValueError):
+            model = vhmm.VariationalGaussianHMM(
```
anntzer, line 318: Instead of repeatedly setting up the same hmm except for the wrong attribute, perhaps factor that out in a new_hmm() method (as exists in other tests) which already sets all attributes to valid values, and just overwrite whatever needs to be in the pytest.raises() block?

anntzer (approved):
Minor nit, otherwise looks reasonable.

lib/hmmlearn/tests/test_variational_gaussian.py
blckmaxima: Great idea, thanks.

anntzer (approved):

anntzer:
Thanks!
I additionally pushed a couple more cleanups (not necessarily directly linked to your PRs, just things I noticed while moving all the code around).

blckmaxima:
Great!

PR #501, opened 2023-01-05
Allow setting n_features for Categorical and VariationalCategorical HMMs

blckmaxima:
I'm porting some code from my thesis to cluster HMMs.  In that algorithm I train a model on each sequence - some sequences may not contain the full set of emissions symbols, so it is necessary to specify the shape of the emissions distribution during HMM setup.

### DIFF
--- a/lib/hmmlearn/_emissions.py
+++ b/lib/hmmlearn/_emissions.py
@@ -19,7 +19,7 @@ class BaseCategoricalHMM(_AbstractHMM):
             raise ValueError("Symbols should be integers")
         if X.min() < 0:
             raise ValueError("Symbols should be nonnegative")
-        if hasattr(self, "n_features"):
+        if self.n_features is not None:
             if self.n_features - 1 < X.max():
                 raise ValueError(
                     f"Largest symbol is {X.max()} but the model only emits "
--- a/lib/hmmlearn/hmm.py
+++ b/lib/hmmlearn/hmm.py
@@ -78,9 +78,9 @@ class CategoricalHMM(_emissions.BaseCategoricalHMM, BaseHMM):
 
     def __init__(self, n_components=1, startprob_prior=1.0,
                  transmat_prior=1.0, *, emissionprob_prior=1.0,
-                 algorithm="viterbi", random_state=None,
-                 n_iter=10, tol=1e-2, verbose=False,
-                 params="ste", init_params="ste",
+                 n_features=None, algorithm="viterbi",
+                 random_state=None, n_iter=10, tol=1e-2,
+                 verbose=False, params="ste", init_params="ste",
                  implementation="log"):
         """
         Parameters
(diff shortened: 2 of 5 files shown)


lib/hmmlearn/hmm.py
```diff
@@ -164,12 +169,13 @@ def _check(self):
         super()._check()
 
         self.emissionprob_ = np.atleast_2d(self.emissionprob_)
-        n_features = getattr(self, "n_features", self.emissionprob_.shape[1])
-        if self.emissionprob_.shape != (self.n_components, n_features):
+        if not self.n_features:
```
anntzer, line 172: `if self.n_features is None` because explicitly passing n_features=0 seems more like a programming error which should raise rather than something that should be silently overwritten.

lib/hmmlearn/tests/test_variational_categorical.py
```diff
@@ -99,6 +111,7 @@ def test_n_features(self, implementation):
             sequences = sequences.astype(float)
             assert_log_likelihood_increasing(model, sequences, lengths, 10)
 
+
```
anntzer, line 114: pep8: one blank line between methods, not two

lib/hmmlearn/vhmm.py
```diff
@@ -176,12 +181,13 @@ def _check(self):
             raise ValueError(
                 "emissionprob_prior_ and emissionprob_posterior_must"
                 "have shape (n_components, n_features)")
-        n_features = getattr(self, "n_features",
-                             self.emissionprob_posterior_.shape[1])
+        if not self.n_features:
```
anntzer, line 184: `if self.n_features is None:`

anntzer:
Just a few nits.

lib/hmmlearn/hmm.py
blckmaxima: :+1:

lib/hmmlearn/_emissions.py
```diff
@@ -19,7 +19,7 @@ def _check_and_set_n_features(self, X):
             raise ValueError("Symbols should be integers")
         if X.min() < 0:
             raise ValueError("Symbols should be nonnegative")
-        if hasattr(self, "n_features"):
+        if hasattr(self, "n_features") and self.n_features is not None:
```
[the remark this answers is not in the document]
blckmaxima: Yes, good catch.

lib/hmmlearn/tests/test_variational_categorical.py
blckmaxima: thanks.

lib/hmmlearn/vhmm.py
blckmaxima: yup!

blckmaxima:
Thanks - updated...

anntzer:
Thanks :)

PR #502, opened 2023-01-07
Correct the model-selection example

blckmaxima:
@anntzer is there a way to see the produced html docs before merging? I poked around the ci/cd a bit, but it was not immediately obvious.  I'm attempting to make the example output reproducible and show a clear minimum for the aic/bic at 4 states.

### DIFF
--- a/examples/plot_gaussian_model_selection.py
+++ b/examples/plot_gaussian_model_selection.py
@@ -16,8 +16,11 @@ of components for a model.
 """
 import numpy as np
 import matplotlib.pyplot as plt
+from sklearn.utils import check_random_state
 
 from hmmlearn.hmm import GaussianHMM
+
+rs = check_random_state(546)
 # %%
 # Our model to generate sample data from:
 model = GaussianHMM(4, init_params="")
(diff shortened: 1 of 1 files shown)


anntzer:
The docs are getting built on each PR in readthedocs, but currently things are associated with my personal readthedocs account and there seems to be some issues with RTD being authorized to push notifications to the hmmlearn org :/  Probably will need some fiddling...
I note that here the plain LL with 6 components is actually worse than the one with 5 components, which I guess suggests that 100 iterations is actually not enough for convergence?
Perhaps also consider shortening the number of samples (is 1000 enough?)

blckmaxima:
Understood, Thank you!  I reduced the data-size, and decreased the tolerance, and received a very-nice looking plot where LL keeps increasing, and a sharp minimum at 4 for the aic/bic.  Does this match what you see  in read-the-docs? Otherwise, I need to work on reproducibility.
sphx_glr_plot_gaussian_model_selection_001

anntzer:
Yup, this works now, thanks!

PR #506, opened 2023-02-06
Initialize startprob/transmat of VI Hmms with Dirichlet sampled estimates, similar to EM HMM.

blckmaxima:


anntzer:
Seems generally reasonable (you need to rebase on master, to which I just force-pushed because I accidentally broke it earlier, sorry), but this severely lengthens the variational part of the test suite (by ~1min here), which is already quite long; can you do something about this?

blckmaxima:
Yeah, let me look at that!

blckmaxima:
I'll look at the failure a bit later...

anntzer:
Did you rebase on main?

blckmaxima:
Yes,  the failure is related to a change I made.

lib/hmmlearn/tests/__init__.py
```diff
@@ -85,3 +85,12 @@ def compare_variational_and_em_models(variational, em, sequences, lengths):
...
     assert np.all(em_obs == vi_obs)
     assert np.all(em_states == vi_states)
+
+
+def vi_uniform_startprob_and_transmat(model, lengths):
+    nc = model.n_components
+    model.startprob_prior_ = np.array([1/nc]*nc)
+    model.startprob_posterior_ = np.array([1/nc]*nc) * len(lengths)
+    model.transmat_prior_ = np.full((nc,nc), 1/nc)
+    model.transmat_posterior_ = np.full((nc,nc), 1/nc) * sum(lengths)
```
anntzer, line 95: At least add a comma in (nc, nc).  I would also rewrite the first two lines as np.full(nc, 1/nc) for consistency with the ones below (`np.array([1/nc]*nc) * len(lengths)` also reads slightly confusingly because the two `*` next to one another have different semantics) but don't feel extremely strongly about it.

anntzer:
Just a minor nit, otherwise LGTM :)

blckmaxima:
Thanks!

PR #507, opened 2023-02-13
Fix gaussian vhmm

blckmaxima:
Addressing subtle issues with the Variational Gaussian HMM

### DIFF
--- a/lib/hmmlearn/vhmm.py
+++ b/lib/hmmlearn/vhmm.py
@@ -647,21 +647,24 @@ class VariationalGaussianHMM(BaseGaussianHMM, VariationalBaseHMM):
         # In general, things are neater if we pretend the covariance is
         # full / tied.  Or, we could treat each case separately, and reduce
         # the number of operations. That's left for the future :-)
-        term1 = np.zeros_like(self.dof_posterior_, dtype=float)
-        for d in range(1, self.n_features+1):
-            term1 += special.digamma(.5 * self.dof_posterior_ + 1 - d)
+        nf = self.n_features
+
+        term1 = special.digamma(
+            .5 * (self.dof_posterior_ - np.arange(0, nf)[:, None])
+        ).sum(axis=0)
+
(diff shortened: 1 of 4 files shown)


blckmaxima:
I am expecting this to fail, and am tracking down a final bug...

blckmaxima:
And yes, I'll be deleting some `print()` before merging :-p

blckmaxima:
@anntzer What are your thoughts on the change I made in `base.py` about "strict".    In testing this week, the best I can understand is the lower bound does not increase due  to floating point rounding error. 

Another option is to actually back out the notion of "strict", and let end-users validate the model training history themselves.

anntzer:
We used not to check at all (AFAICT) and you introduced the check in the vhmm PR.  I guess the check is useful, but allowing floating point slop is reasonable (sqrt(eps) seems like a reasonable allowance on general grounds), and choosing e.g. to warn instead of raising would probably be good too.  I would then get rid of the "strict" API and let users rely on the stdlib warnings API to control the warnings.

blckmaxima:
Good call on backing out the `strict` parameter I added.

blckmaxima:
@anntzer This should be good to go.

lib/hmmlearn/vhmm.py
```diff
@@ -784,7 +787,7 @@ def _compute_lower_bound(self, log_prob):
         if self.covariance_type != "full":
             scale_posterior_ = fill_covars(self.scale_posterior_,
                     self.covariance_type, self.n_components, self.n_features)
-            scale_prior_ = fill_covars(self.scale_posterior_,
+            scale_prior_ = fill_covars(self.scale_prior_,
```
anntzer, line 790: Is this effectively the only material change?  (I trust you on the formulas...)

blckmaxima: I have a follow-on MR that will then reuse most of this in a GMM, and I've been comparing the formulas across a few different papers.

PR #515, opened 2023-03-30
fix `n_features` for Bernoulli distribution

tuix:
In the Multinomial example the number of features needed for a Bernoulli distribution is specified 1 where it should be 2


anntzer (approved):

anntzer:
Thanks!

PR #516, opened 2023-04-03
GMM-HMM: Simplifiy / Explicitly refer to literature for implementation

blckmaxima:
Re-implement GMMHMM based upon Newer Reference 
   Watanabe, Shinji, and Jen-Tzung Chien. Bayesian Speech and Language
    Processing. Cambridge University Press, 2015.
which appears simpler than the current implementation. This will make the VariationalGMMHMM Easier to implement later on.


blckmaxima:
This changes the implementation, so I'm aware that we should be cautious before merging.  However, I like that the implementation clearly follows a reference. I've called out in the code where I didn't have any references (nor where I could find a reason for what was going on).

blckmaxima:
@anntzer Thinking of merging this (and the Variational GMM work), and making a release in the coming weeks... thoughts?

lib/hmmlearn/_emissions.py
```diff
@@ -263,33 +262,23 @@ def _accumulate_sufficient_statistics(self, stats, X, lattice,
             post_comp_mix = post_comp[:, :, None] * post_mix
 
         stats['post_mix_sum'] += post_comp_mix.sum(axis=0)
-        stats['post_sum'] += post_comp.sum(axis=0)
-
         if 'm' in self.params:  # means stats
             stats['m_n'] += np.einsum('ijk,il->jkl', post_comp_mix, X)
 
         if 'c' in self.params:  # covariance stats
-            centered = X[:, None, None, :] - self.means_
```
anntzer, line 272: From a quick skimming, if I understand correctly (I didn't actually check the formulas), this is something akin replacing var(X) = <(X - <X>)^2> by var(X) = <X^2> - <X>^2?  (you don't center the X's anymore)?  Do you know if one approach is better/faster/more numerically accurate than the other?

Other than that I guess this looks reasonable?  (Again, without actually checking the math...)

blckmaxima: Good question, that I don't immediately know the answer to.  Let me look at something.

PR #531, opened 2023-08-31
Account for different ordering of results in the  Gaussian Tests

blckmaxima:


anntzer (commented):
FWIW, looks good to me.

PR #562, opened 2024-09-09
Update CI re: deprecated actions.

anntzer:


blckmaxima:
@anntzer For this, do you want to wait until sklearn is updated for python 3.13?

anntzer:
Let's just skip the tests on Windows + python 3.13; the wheels don't actually link to it so I have no doubt that they will be correct if the other tests pass.

anntzer:
I think now it's just a matter of adding a changelog entry, tagging, and uploading everything to PyPI.  @blckmaxima Can you take care of that?

PR #564, opened 2024-10-04
hack for GIL bug

mr-c:
In Debian, we need this patch to allow the tests to pass, otherwise 202 of them fail.

Here's a full log. I can also create a `Dockerfile` demonstrating the bug, for the curious.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1082695

Sample error
```
    def _fit_scaling(self, X):
        frameprob = self._compute_subnorm_likelihood(X)
>       logprob, fwdlattice, scaling_factors = _hmmc.forward_scaling(
            self.startprob_subnorm_, self.transmat_subnorm_, frameprob)
E       RuntimeError: pybind11::handle::inc_ref() PyGILState_Check() failure.

hmmlearn/base.py:1103: RuntimeError
----------------------------- Captured stderr call -----------------------------
pybind11::handle::inc_ref() is being called while the GIL is either not held or invalid. Please see https://pybind11.readthedocs.io/en/stable/advanced/misc.html#common-sources-of-global-interpreter-lock-errors for debugging advice.
If you are convinced there is no bug in your code, you can #define PYBIND11_NO_ASSERT_GIL_HELD_INCREF_DECREF to disable this check. In that case you have to ensure this #define is consistently used for all translation units linked into a given pybind11 extension, otherwise there will be ODR violations. The failing pybind11::handle::inc_ref() call was triggered on a numpy.ndarray object.
```

### DIFF
--- a/ext/_hmmc.cpp
+++ b/ext/_hmmc.cpp
@@ -1,3 +1,4 @@
+#define PYBIND11_NO_ASSERT_GIL_HELD_INCREF_DECREF
 #include <pybind11/pybind11.h>
 #include <pybind11/numpy.h>
 #include <cfenv>


mr-c:
> I would assume that this is actually the marker of a real GIL-related bug that needs to be fixed.

Yes, that is also my impression

anntzer:
@mr-c can you check whether #566 fixes the issue for you?

mr-c:
@anntzer I applied https://github.com/hmmlearn/hmmlearn/pull/566 to the Debian package instead of what I have here and it worked!

PR #565, opened 2024-10-04
add progress bar when fitting model

StevenChen16:


anntzer:
If we really want this (which is honestly not so clear to me) we should probably copy the API e.g. of emcee (https://emcee.readthedocs.io/en/stable/user/sampler/#emcee.EnsembleSampler.sample); see e.g. how they implement progress and progress_kwargs.

PR #566, opened 2024-10-06
Reacquire GIL before constructing tuples containing numpy arrays.

anntzer:
... as explicitly required by pybind11.  Compile with `#define PYBIND11_ASSERT_GIL_HELD_INCREF_DECREF` to observe GIL assertion failures in absence of this patch.

Closes (and supersedes) #564.

(See item regarding "pybind11 objects that are members of other C++ structures" in https://pybind11.readthedocs.io/en/stable/advanced/misc.html#common-sources-of-global-interpreter-lock-errors)


mr-c (approved):
The fix is confirmed from the Debian side

PR #581, opened 2026-03-27
bump minimum manylinux version to 2.28, to match numpy 2.3.0+

mr-c:
See also https://github.com/numpy/numpy/pull/28436


mr-c:
This was necessary to test locally for me; without this change then testing takes an extremely long time as the pydata stack is no longer publishing older manylinux binary wheels; and eventually scipy fails to compile for manylinux_2_17 due to the lack of openblas library.

If you don't mind, I can add a commit to move the cibuildwheel configuration from the GitHub action to `pyproject.toml`, which will make it easier for others to run cibuildwheel. Otherwise they have to translate these settings to environment variables by hand: `CIBW_MANYLINUX_AARCH64_IMAGE=manylinux_2_28 CIBW_MANYLINUX_X86_64_IMAGE=manylinux_2_28 CIBW_BUILD='cp313-*' CIBW_SKIP='*-musllinux_*' CIBW_ARCHS_LINUX='x86_64 aarch64' CIBW_TEST_REQUIRES=pytest  CIBW_TEST_COMMAND='python -I -mpytest --pyargs hmmlearn.tests {project}/doc' cibuildwheel` versus just `cibuildwheel` + any overrides they desire.

